PR #74, opened 2023-03-27
first draft for questionnaire generation

kaligrafy:


samuel-duhaime:
I think we could close this pull request also, to focus on Evolution-generator. 
If we need to implement functionality like Generic Widgets, we should add it to the to-do list of Generator #394 .

greenscientist:
Closing this, we have a new path forward

PR #75, opened 2023-03-29
mode types and helper to common

kaligrafy:
This allows the helpers and types to be used both in backend and frontend, and also use the interview types for the questionnaire config instead of creating separated config types


tahini (commented):
Sounds good, but the frontend user will have to be moved first, then the version of chaire-lib updated in evolution.

packages/evolution-common/package.json
```diff
@@ -20,6 +20,9 @@
     "format": "prettier-eslint $PWD/'src/**/*.{ts,tsx}' --write"
   },
   "dependencies": {
+    "@fortawesome/fontawesome-svg-core": "^1.2.32",
+    "@fortawesome/free-solid-svg-icons": "^5.15.1",
```
tahini, line 24: Except for fontawesome-svg-core, do you need the other 2 packages? If not, please remove from dependencies

packages/evolution-common/src/utils/labels.TEST
```diff
@@ -0,0 +1,34 @@
+import {
+    TranslationText
+} from '../types/types';
+
+export const sliderLabels: { [key: string]: TranslationText[] } = {
```
tahini, line 5: I don't think this file should be committed?

packages/evolution-common/src/utils/validations.TEST
```diff
@@ -0,0 +1,42 @@
+import { Validation } from '../types/types';
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
+
+export const generic: { [key: string]: ((value: any) => Validation) } = {
```
tahini, line 4: Neither this one

tahini (changes requested):
Some files should not be committed I think. Or if they have a reason to be there, add a proper extension and comment if possible.

tahini (approved):
Ok, but next time, the transition submodule update will have to be in a separate commit/PR

PR #76, opened 2023-03-30
add new types for travel surveys

kaligrafy:
includes tests for each new type and updates of custom generic types

### DIFF
--- a/packages/evolution-common/src/services/interviews/__tests__/interview.test.ts
+++ b/packages/evolution-common/src/services/interviews/__tests__/interview.test.ts
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2023, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { InterviewResponses, SectionActionStatus } from '../interview';
+import { Household } from '../../interviewObjects/Household';
+import { Place } from '../../interviewObjects/Place';
+import { Vehicle } from '../../interviewObjects/Vehicle';
+import { Person } from '../../interviewObjects/Person';
+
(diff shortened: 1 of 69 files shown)


packages/evolution-common/src/services/interviews/__tests__/interview.test.ts
```diff
@@ -0,0 +1,121 @@
...
+type CustomPerson = { customPersonAttribute: string };
+type CustomPlace = { customPlaceAttribute: string };
+type CustomVehicle = { customVehicleAttribute: string };
+type CustomVisitedPlace = { customVisitedPlaceAttribute: string };
+type CustomTrip = { customTripAttribute: string };
+type CustomSegment = { customSegmentAttribute: string };
+
+describe('InterviewResponses type tests', () => {
+    test('Valid InterviewResponses values', () => {
+        const interviewResponses: InterviewResponses<CustomSurvey, CustomHousehold, CustomPerson, CustomPlace, CustomVehicle, CustomVisitedPlace, CustomTrip, CustomSegment> = {
```
tahini, line 24: Most of the added tests are trivial, but it does make sure that this object compiles. Maybe add a comment explaining that this tests more the compilation of this object than behavior. And everywhere such tests exist

tahini (commented):
This is showing the limits of the current architecture. When it was just person, household and home and that all foreseeable questionnaire used those, it was fine. But adding vehicles, trips segments and visited places, some of which are optional (like vehicle, I haven't seen it in a questionnaire yet), I'm thinking this is wrong... I'll have to sleep on it. We need to offer questionnaire developers all the required flexibility, while allowing strong typing and keeping it simple.

For vehicles, household, etc, composition may be preferable to generics, but then how do we define the types for validations? To be continued....

Issue #79, opened 2023-04-19
survey: Implement the participant login, separate from user login


tahini:
This is very similar to the user login, but it will look in a different table for the users, and thus use different server routes.

We should maximize code re-use for both flows.

Issue #78, opened 2023-04-19
survey: Implement a participant table separate from users


tahini:
Participants is a very different concept than users and should not have the same rights/permissions. If they are all in the same table, someone administering users could accidentally give permissions to a simple participant, which may cause security issues.

Individual participants filling interview from the web should be in a separate table. There is a one participant <-> one interview rule. Participants will be based on the current User table

Issue #51, opened 2022-10-07
evolution: Implement a separate application for participants and other users


tahini:
Independent of where the users vs participants are stored, a separate "user" login page can be configured the same way as the main application's login (see the auth configuration options).

- [ ] In the backend, allow to specify which `auth` options to use for the various login pages
- [ ] Since auth options are different, it would make sense to have them served by different auth routes, so we should be able to set a prefix to the routes
- [ ] In the frontend, all pages regarding an authentication method (for example login/register/forgotPwd/reset/etc for localLogin) should have the URL links configurable through the props.

PR #82, opened 2023-05-01
Split the app in a participant and admin app

tahini:
** Breaking change for surveys **

This is a  big PR, which solves 3 issues: #51, #78 and #79. Each individual commit is standalone, but they break the application's functionalities until they are all there, so there are part of a single PR.

* It uses the participants table to link to the interviews instead of the users table (current users with interviews are copied to the participant table so current survey apps still work)
* There's a separate participant authentication model, which can use the same authentication methods as the user model, but uses the participants table instead of the users
* A participant app has a login page that uses the participant authentication model and can only display the participant's interview. It has no other rights.
* An admin app has a login page using the users authentication model. Users do not have an interview, their default home page displays their allowed pages.

When this PR is merged, all current surveys will need to be updated to use the latest evolution. See instructions in the individual commits for how to update


kaligrafy (approved):

greenscientist:
This probably need a big update to the README of the app. Just having the configuration information in the commit message is not enough.

greenscientist (commented):
Kind of wondering if we should add the login to the partipant table directly or to a side table instead.  (Unless the participant table is only used for login, then fine.)

tahini:
@greenscientist you're right about the readme, totally forgot about it! 

As for the participant table, all those fields are for the login support of various methods (facebook, google, local login, anonymous, with confirmation, reset password, etc), mostly copy of the users table minus the user specific fields (is_admin, preferences, etc). Actually, the participant table is a side table for the sv_interviews.

tahini:
@greenscientist just added the README

tahini:
I'll merge so it can be used by projects right away. If anything is not clear or not completely working, we can fix later.

PR #103, opened 2023-05-16
Allow forcing string input to be lowercase or uppercase

davidmurray:
Implements necessary features to fix issue https://github.com/chairemobilite/od_mtl_2023/issues/400

Thanks!

### DIFF
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -37,6 +37,7 @@ export type InputStringType<CustomSurvey, CustomHousehold, CustomHome, CustomPer
     maxLength?: number;
     datatype?: 'string' | 'integer' | 'float';
     size?: 'large' | 'small' | 'medium';
+    textTransform?: 'none' | 'capitalize' | 'uppercase' | 'lowercase';
 };
 
 export type InputTextType<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> = {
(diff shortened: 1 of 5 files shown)


packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -37,6 +37,7 @@ export type InputStringType<CustomSurvey, CustomHousehold, CustomHome, CustomPer
     maxLength?: number;
     datatype?: 'string' | 'integer' | 'float';
     size?: 'large' | 'small' | 'medium';
+    textTransform?: string;
```
tahini, line 40: There might be a better type than string for this, like 'none' | 'capitalize' | 'uppercase' | 'lowercase'. I think that can be sufficient for us.

tahini (changes requested):
Thanks, this is nice! Just a small comment to make the type safer. Also, this affects only the display of the value, not the value itself, which remains as entered by the user (so maybe lowercase). But I'm fine with this.

packages/evolution-common/src/services/widgets/WidgetConfig.ts
davidmurray: Fixed in commit 87c2e27 🙂

packages/evolution-frontend/src/components/inputs/__tests__/InputString.test.tsx
```diff
@@ -32,6 +32,7 @@ test('Should correctly render InputString with all parameters', () =>{
         datatype: 'string' as const,
         maxLength: 20,
         defaultValue: 'test',
+        textTransform: 'lowercase',
```
tahini, line 35: To have the test pass, here you should put `'lowercase' as const` to avoid the type being `string`

PR #112, opened 2023-05-24
Add a callback function to filter geocoding results when calculating new map bounds

davidmurray:
* This is necessary in order to fix issue https://github.com/chairemobilite/od_mtl_2023/issues/416
* This callback function can be used to exclude points that are really far from the survey area. These points will still be displayed on the map, but at least the map will stay centered around the survey area.


kaligrafy:
We could also add a maxMapBounds which would be a config for the survey. That way, we just need to verify that the calculated bounds has at least one corner inside the maxMapBound. If not, use the max map bound.

packages/evolution-frontend/src/components/inputs/maps/google/InputMapGoogle.tsx
```diff
@@ -135,10 +136,13 @@ const InputMapGoogle: React.FunctionComponent<InputGoogleMapPointProps> = (props
     React.useEffect(() => {
         if (map && props.markers.length >= 1) {
             const mapBounds = new google.maps.LatLngBounds();
-            props.markers.forEach((marker) =>
+            const points = props.markers.map((marker) => marker.position.geometry) as [GeoJSON.Point];
```
tahini, line 139: I think the syntax here would be `GeoJSON.Point[]` for an array of points. `[GeoJSON.Point]` is an array with a single point.

tahini:
I'm reviewing the code, but I'll let @kaligrafy review if this is the desired approach for the problem.

davidmurray:
> Is it possible to also accept a max bound? (min/max lat and lon)

I just pushed a new method (see commit [cd65c0f](https://github.com/chairemobilite/evolution/pull/112/commits/cd65c0f95f60dcd57b32a05beace073603e62093)). Is this what you were suggesting?

tahini:
@davidmurray Before committing, you can run `yarn format` on the workspaces that you touched, or just from the root of the repo. This will fix many formatting and lint issues. The code-lint test does not catch all formatting problems unfortunately, so sometimes, we have commits where we only ran `yarn format` on the repo. But the format command will catch pretty much all linting errors and sometimes auto-fix them.

kaligrafy:
That's good, but if the only markers found are outside the max bound, we should use the markers bounds. It is only when there are at least one marker inside the maxbound

davidmurray:
> That's good, but if the only markers found are outside the max bound, we should use the markers bounds. It is only when there are at least one marker inside the maxbound

Fixed (see latest commit).

Now it works like this:
image
image
image

So it correctly filters out results in the USA (i.e. outside max bounds)... yet it keeps results that are in Ottawa (which are also outside the max bounds). hmmm...

kaligrafy:
Seems correct

tahini (approved):
Let's try this live and see if it works well.

PR #131, opened 2023-06-26
inputSelect: Add line separator for groups without label

### LINKED ISSUE #130: inputSelect: add thin line to split optgroup if no label
Required to fix https://github.com/chairemobilite/od_mtl_2023/issues/504

If the optgroup has no label, instead, there should be a thin separator for the optgroups instead of actual optgroups

tahini:
Fixes #130

Select choices can be separated in groups with group labels. If the label is set to an empty string, instead of adding an indented optgroup with an empty label, add a disabled horizontal line option to separate the group from the rest.

Inpired from this stackoverflow answer: https://stackoverflow.com/questions/899148/html-select-option-separator

Update the car driver's parking question to split in group with empty label. And re-activate the dwelling type question, for an example with groups with labels.

### DIFF
--- a/example/demo_survey/src/survey/sections.ts
+++ b/example/demo_survey/src/survey/sections.ts
@@ -27,7 +27,7 @@ const homeWidgets = [
   'homeRegion',
   'homeCountry',
   'homePostalCode',
-  //'homeDwellingType',
+  'homeDwellingType',
   'homeGeography'
 ];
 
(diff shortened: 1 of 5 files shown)


tahini:
Here's a screenshot showing what the select with empty labels look like:

Capture d’écran du 2023-06-26 16-42-07

kaligrafy (approved):

PR #132, opened 2023-06-27
Add a footer to a survey page

davidmurray:
The footer will be populated using the contents of the `survey:footer` translatable key

image

**TODO**: Make it so that the footer sticks to the bottom of the page. My CSS skills are not good enough for this and I will focus on more pressing issues instead

### DIFF
--- a/packages/evolution-legacy/src/components/survey/Survey.js
+++ b/packages/evolution-legacy/src/components/survey/Survey.js
@@ -204,6 +204,7 @@ export class Survey extends React.Component {
                   submitted                   = {this.props.submitted}
                 />
               </form>
+              <div dangerouslySetInnerHTML={{__html: this.props.t(['survey:footer', 'main:footer']) }} />
             </div>
           </div>
         );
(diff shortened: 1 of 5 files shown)


packages/evolution-legacy/src/components/survey/Survey.js
```diff
@@ -204,6 +204,7 @@ export class Survey extends React.Component {
                   submitted                   = {this.props.submitted}
                 />
               </form>
+              <div dangerouslySetInnerHTML={{__html: this.props.t(`survey:footer`)}} />
```
tahini, line 207: To avoid surveys who do not have this string defined to have the text `survey:footer` at the bottom, you can do `props.t(['survey:footer', 'main:footer'])`, define the `footer` string in evolution's `main.json` file in the locale (it can be empty).

For the example in the demo_survey, you could use more generic links than the artm's own (this project is open source and could be used by anyone in the world). For example, a link to evolution's github page!

tahini (changes requested):

packages/evolution-legacy/src/components/survey/Survey.js
davidmurray: Good points! Fixed :)

davidmurray:
image

tahini (approved):

PR #134, opened 2023-06-28
logging: use a logging middleware to log user accesses to interviews

### LINKED ISSUE #43: evolution: Find a solution to keep track of users who edited an interview
The quick&easy solution is to add a field in responses with an array of user ids who edited the interview. But this exposes the IDs of users with higher privileges than participants. That is not good. It should be handled only server side and never make it to the client.

tahini:
Fixes #43 

Add a database table, which tracks users who access a given interview, counting the number of edits done in the interview, with start and update timestamps

The express routes to open and update the interview use an additional middleware, which saves the user access to the interview. In the admin application, the middleware saves the data to the table described above. In the participant applicaiton, it does nothing.

The previous _editingUsers field in the responses is dropped and the statEditingUsers function is updated accordingly.

### DIFF
--- a/packages/evolution-backend/src/services/logging/queryLoggingMiddleware.ts
+++ b/packages/evolution-backend/src/services/logging/queryLoggingMiddleware.ts
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2023, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { UserAttributes } from 'chaire-lib-backend/lib/services/users/user';
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
+import { NextFunction, Request, Response } from 'express';
+import interviewsAccessesDbQueries from '../../models/interviewsAccesses.db.queries';
+
+export type InterviewLoggingMiddlewares = {
+    openingInterview: (validationMode: boolean) => (req: Request, res: Response, next: NextFunction) => void;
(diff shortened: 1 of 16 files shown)


tahini:
@greenscientist You may want to take a look at this PR? The logging middleware can eventually be used to do more stuff... like save the actual fields touched by the update, send data to an external logging infra, etc.

kaligrafy (approved):

tahini:
I'm mergine, we need this for od_mtl, we can improve later

packages/evolution-backend/src/services/logging/queryLoggingMiddleware.ts
```diff
@@ -0,0 +1,49 @@
...
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
+import { NextFunction, Request, Response } from 'express';
+import interviewsAccessesDbQueries from '../../models/interviewsAccesses.db.queries';
+
+export type InterviewLoggingMiddlewares = {
+    openingInterview: (validationMode: boolean) => (req: Request, res: Response, next: NextFunction) => void;
+    updatingInterview: (validationMode: boolean) => (req: Request, res: Response, next: NextFunction) => void;
+};
+
+const openingInterview = (validationMode: boolean) => (req: Request, _res: Response, next: NextFunction) => {
```
greenscientist, line 17: What is the validationMode? What does true/false means here? It's no obvious, would need comments

greenscientist (commented):
SOrry, late reading this. Adding more info can be put in a new issue.

greenscientist:
I don't think we should be adding too much stuff to this "loggingMiddleWare". I think we should keep it quite access specific. We could do something else for other type of logging and we will want to stick to a more standard solution.

PR #147, opened 2023-07-11
InputMapFindPlace: display a warning in red if the geocoding query is too imprecise

davidmurray:
If there is only a single result, we check if it's precise enough using the location type information returned by the geocoding API. The Google Maps geocoding API returns an array of location types for each result. For a result, if all of its location types are deemed too imprecise, then this result is considered to be imprecise and an error is shown to the user.

--
@tahini  @kaligrafy would like both of you to review if possible!

Here are some example results:
image
image
image
image

Unfortunately we can still choose large areas. For example, Île Notre-Dame is tagged as 'natural_feature', which could be really big or really small so we can't do much about this.
image

Thanks!


davidmurray:
tests are failing... I will investigate why

davidmurray:
@kaligrafy Address location types are saved in the interview object:
```json
"1a56cbf8-7ec3-4ae4-83f1-bc5ff3182d3c":{
                  "_sequence":3,
                  "_uuid":"1a56cbf8-7ec3-4ae4-83f1-bc5ff3182d3c",
                  "activity":"shopping",
                  "name":"parc jeanne mance",
                  "geography":{
                     "type":"Feature",
                     "properties":{
                        "isGeocodingImprecise":false,
                        "lastAction":"findPlace",
                        "geocodingQueryString":"parc jeanne mance",
                        "geocodingResultsData":{
                           "formatted_address":"4422 Esplanade Ave, Montreal, Quebec H2W 2N4, Canada",
                           "place_id":"ChIJPxWAhzIayUwRcPG_kiLoJzc",
                           "types":[
                              "park",
                              "point_of_interest",
                              "establishment"
                           ]
                        }
                     },
                     "geometry":{
                        "type":"Point",
                        "coordinates":[
                           -73.58697459999999,
                           45.5177162
                        ]
                     },
                     "id":"ChIJPxWAhzIayUwRcPG_kiLoJzc"
                  },
                  "arrivalTime":55500,
                  "nextPlaceCategory":"wentBackHome",
                  "departureTime":58200,
                  "_isNew":false
               }
```

kaligrafy:
OK, so you can add validations for the types attribute

kaligrafy:
C'est super en tout cas!

davidmurray:
> IL faudrait pas oublier d'arrondir toutes les coordonnées à 5 décimales

oui! je voulais faire ça dans une autre PR

kaligrafy:
oui, et probablement dans evolution et pas ici

davidmurray:
OK, I added a new unit test and fixed the broken one(s)

example/demo_survey/src/survey/widgets/visitedPlaces.tsx
```diff
@@ -808,6 +808,23 @@ export const visitedPlaceGeography = {
   geocodingQueryString: function (interview, path) {
     return surveyHelperNew.formatGeocodingQueryStringFromMultipleFields([surveyHelper.get(interview, path, null, '../name')]);
   },
+  invalidGeocodingResultTypes: [
```
tahini, line 811: Should these be in the survey itself? Or should they be part of the map. This is Google maps API specific, so, if it goes in the platform, it should be in some class with Google in its name. Maybe it's easier to put here though, at least for now

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
```diff
@@ -228,6 +263,11 @@ export class InputMapFindPlace<CustomSurvey, CustomHousehold, CustomHome, Custom
             selectedPlace.properties = { ...selectedPlaceProperties };
             selectedPlace.properties.lastAction = 'findPlace';
             selectedPlace.properties.geocodingQueryString = this.state.geocodingQueryString;
+            selectedPlace.properties.geocodingResultsData = {
+                formatted_address: placeData.formatted_address,
+                place_id: placeData.place_id,
```
tahini, line 268: Are we allowed to save this? But this is the main InputMapFindPlace class, so not google specific. If we are not allowed, this is not where we should take care of this, but make sure that in the present case with google, we respect the license

tahini (commented):
Some questions before approval. I love that there's unit tests for this!

example/demo_survey/src/survey/widgets/visitedPlaces.tsx
davidmurray: Yeah, I had the same question myself.
On one hand, it's Google Maps specific, so it could be in evolution directly.
On the other hand, I think it's nice to have the ability in the survey to decide on a per question basis what is considered as "imprecise" or not. Having the types directly in the survey makes this very easy. 

Could leave this for now and see later if we decide to improve this feature

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
davidmurray: https://cloud.google.com/maps-platform/terms/maps-service-terms

> 5. Places API
> 
> The following terms apply only to the Places API:
> ...
> 5.4 Caching. Customer can temporarily cache **latitude (lat) and longitude (lng) values from the Places API for up to 30 consecutive calendar days**, after which Customer must delete the cached latitude and longitude values. Customer can cache Places API Place ID (place_id) values, in accordance with the [Places API Policies](https://developers.google.com/maps/documentation/places/web-service/policies).

https://developers.google.com/maps/documentation/places/web-service/policies

> Note that **the place ID, used to uniquely identify a place, is exempt from the caching restriction.** You can therefore store place ID values indefinitely. The place ID is returned in the place_id field in Places API responses.

https://cloud.google.com/maps-platform/terms
> 
> 3.2.3 Restrictions Against Misusing the Services.
> 
> (a)  No Scraping. Customer will not export, extract, or otherwise scrape Google Maps Content for use outside the Services. For example, Customer will not: (i) pre-fetch, index, store, reshare, or rehost Google Maps Content outside the services; (ii) bulk download Google Maps tiles, Street View images, geocodes, directions, distance matrix results, roads information, places information, elevation values, and time zone details; **(iii) copy and save business names, addresses, or user reviews;** or (iv) use Google Maps Content with text-to-speech services.

Donc, si j'ai bien compris:
1. Place ID: OK ✅ 
2. Lat Lon: 30 jours seulement? ❓ Est-ce que ceci pourrait être un problème pour l'enquête OD en général?
3. address: Non ❌ , donc on peut pas enregistrer le type d'adresse non plus.

cc @kaligrafy

kaligrafy: On viole la license depuis le début, comme tout le monde. Pour régler ça, faudra vraiment que tous les partenaires financent des ressources pour OpenStreetMap au plus vite. Je ne fais plus d'enquête avec Google à l'avenir!

kaligrafy: Ils ne peuvent pas protéger l'adresse, c'est public, ce n'est pas protégé. Le lat lon associé par Google à l'adresse, lui est protégé. Pour le type de précision, c'est ok si on l'utilise live pour aider l'usager, mais effectivement, pas sur qu'on peut garder ça longtemps et faire quelque chose avec... Mais bon, après on est académique et gouvernmental et il y a le fameux "fair-use" qui te permet de faire ce que tu veux pour le greater good si ce n'est pas commercial, mais ça se discute évidemment...

kaligrafy: Selon le fair-use américain et canadien, si tu es de Poly et que tu fais de la recherche avec les données Google: Fair use s'applique, tu fais pas mal ce que tu veux et tu peux publier aussi. Si tu es ARTM: ben ça dépend, est-ce d'intérêt public, est-ce des nouvelles journalistiques quand tu publies une analyse sur les déplacements dans une journal? Zone grise! Pour ça qu'il faut aller avec OpenStreetMap au plus vite!

tahini (approved):
If you are satisfied with this, I see no problem. Maybe just open an issue to track that we'll need to decide how to deal with imprecise locations eventually, that now it is a question-specific parameter.

davidmurray:
Cool!
Par contre ici je n'ai pas les droits de faire un merge :(
@tahini je te laisse le faire !

tahini (changes requested):
Il y a un conflit sur cette PR

davidmurray:
@tahini  fixed :)

tahini (approved):

PR #149, opened 2023-07-12
Allow rounding lat lon coordinates in InputMapFindPlace

davidmurray:
By default, coordinates are rounded to 5 decimal places. At worst, this is a +/- 1 meter precision loss. https://en.wikipedia.org/wiki/Decimal_degrees#Precision

### DIFF
--- a/packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
+++ b/packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
@@ -127,6 +127,12 @@ export class InputMapFindPlace<CustomSurvey, CustomHousehold, CustomHome, Custom
     }
 
     onValueChange = (feature?: GeoJSON.Feature<GeoJSON.Point, FeatureGeocodedProperties>) => {
+        if (feature) {
+            feature.geometry.coordinates = feature.geometry.coordinates.map((coordinate) => {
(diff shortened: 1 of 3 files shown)


packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
```diff
@@ -127,6 +127,12 @@ export class InputMapFindPlace<CustomSurvey, CustomHousehold, CustomHome, Custom
     }
 
     onValueChange = (feature?: GeoJSON.Feature<GeoJSON.Point, FeatureGeocodedProperties>) => {
+        if (feature) {
+            feature.geometry.coordinates = feature.geometry.coordinates.map((coordinate) => {
+                return Number(coordinate.toFixed(this.props.widgetConfig.coordinatesPrecision ?? 5));
```
tahini, line 132: Is `??` like `||` ?

tahini (approved):
Just a question

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
davidmurray: Almost!
See https://stackoverflow.com/a/61481631/1218712

In this case, I think using `??` is better since. If `coordinatesPrecision` is 0, then `||` would consider it to be falsy and it would take the default value.

PR #151, opened 2023-07-14
Add sorting and display options to a question's answers.

Frost000:
With the use of the new parameters, radio and checkbox type questions can have their answers aligned by column and row and have their answers ordered vertically or horizantally. Use 'alignChoices' to have the options aligned, 'columns' to target a number of columns , 'rows' to target a number of rows, 'alignment' to specify in wich direction to order the choices, 'customAlignmentLengths' to specify the length of each row or column. When using custom lengths in horizontal alignment, the first row must be the longuest.

### DIFF
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -117,7 +117,6 @@ export type InputCheckboxType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     sameLine?: boolean;
     customLabel?: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     datatype?: 'string' | 'integer' | 'float' | 'text';
-    alignChoices?: boolean;
     columns?: number;
     rows?: number;
     alignment?: 'vertical' | 'horizontal' | 'auto';
@@ -144,7 +143,6 @@ export type InputRadioType<CustomSurvey, CustomHousehold, CustomHome, CustomPers
     customLabel?: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     customChoice?: string;
     datatype?: 'string' | 'integer' | 'float' | 'text' | 'boolean';
-    alignChoices?: boolean;
     columns?: number;
     rows?: number;
     alignment?: 'vertical' | 'horizontal' | 'auto';
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -121,7 +121,6 @@ export type InputCheckboxType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     columns?: number;
     rows?: number;
     alignment?: 'vertical' | 'horizontal' | 'auto';
-    customAlignment?: boolean;
     customAlignmentLengths?: number[];
 };
 
@@ -149,7 +148,6 @@ export type InputRadioType<CustomSurvey, CustomHousehold, CustomHome, CustomPers
     columns?: number;
     rows?: number;
     alignment?: 'vertical' | 'horizontal' | 'auto';
-    customAlignment?: boolean;
     customAlignmentLengths?: number[];
 };
 
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -114,10 +114,15 @@ export type InputCheckboxType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     seed?: number;
     shuffle?: boolean;
     addCustom?: boolean;
-    columns?: number;
     sameLine?: boolean;
     customLabel?: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     datatype?: 'string' | 'integer' | 'float' | 'text';
+    alignChoices?: boolean;
+    columns?: number;
+    rows?: number;
+    alignment?: 'vertical' | 'horizontal' | 'auto';
+    customAlignment?: boolean;
+    customAlignmentLengths?: [number];
 };
 
 export type InputRadioType<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> = {
@@ -136,11 +141,16 @@ export type InputRadioType<CustomSurvey, CustomHousehold, CustomHome, CustomPers
     seed?: number;
     shuffle?: boolean;
     addCustom?: boolean;
-    columns?: number;
     sameLine?: boolean;
     customLabel?: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     customChoice?: string;
     datatype?: 'string' | 'integer' | 'float' | 'text' | 'boolean';
+    alignChoices?: boolean;
+    columns?: number;
+    rows?: number;
+    alignment?: 'vertical' | 'horizontal' | 'auto';
+    customAlignment?: boolean;
+    customAlignmentLengths?: number[];
 };
 
 // TODO Could select widget have a custom 'other' field? Like checkbox and radios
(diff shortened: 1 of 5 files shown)


packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -0,0 +1,154 @@
...
+            rowIndex = 0;
+            index = columnIndex;
+            out.push(newColumn);
+            newColumn = [];
+        }
+    }
+
+    return out;
+}
+export { customHorizontal };
```
tahini, line 112: Those functions should remain internal to the file and not exported, so the unit test will need to test only the main sortByParameter function, with the proper values as parameters

packages/evolution-frontend/src/components/inputs/__tests__/InputChoiceSorting.test.ts
```diff
@@ -0,0 +1,66 @@
+import * as inputChoiceSorting from '../InputChoiceSorting';
```
tahini, line 1: Copyright header here too please

packages/evolution-frontend/src/components/inputs/__tests__/InputChoiceSorting.test.ts
```diff
@@ -0,0 +1,66 @@
...
+        expect(result).toEqual([[0,1,2,3,4],[5,6,7,8,9]]);
+    });
+    test('with invalid parameters.', () => {
+        const columns = -1;
+        const alignment = "vertical";
+        const result = inputChoiceSorting._sortByParameters(array, alignment, columns);
+
+        expect(result).toEqual([0,1,2,3,4,5,6,7,8,9]);
+    });
+    test('with custom vertical parameters.', () => {
```
tahini, line 52: For custom, make sure to add some tests with more and less items than custom lengths, to make sure the results are as expected (and document what is expected, for less, you stop at the number of items, for more though... do you repeat the custom orders, or just the last? Or do you ignore the additional items (I think not)?)

packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const _sortByParameters = function (arr: any[], alignment: 'vertical' | 'horizontal' | 'auto', columns?: number, rows?: number, customAlignmentLengths?: number[]) {
```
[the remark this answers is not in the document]
Frost000: I did not type the input because the functions only reorders the content of the array, what is in the array is not important.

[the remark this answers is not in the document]
tahini: In that case, you can make it generic by typing the array to T. This will ensure that callers will receive the same type as what they sent instead of receiving any and not being able to type check anymore

packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -114,10 +114,14 @@ export type InputCheckboxType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     seed?: number;
     shuffle?: boolean;
     addCustom?: boolean;
-    columns?: number;
     sameLine?: boolean;
     customLabel?: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     datatype?: 'string' | 'integer' | 'float' | 'text';
+    alignChoices?: boolean;
```
tahini, line 120: I did not notice the first time I think, but alignChoices is redundant with the presence of either rows and columns. We should not require to set alignChoices if we have columns or rows, just as alignChoices = true does not make sense if the other values are not set. So we should have columns, rows, customAlignmentLength. And the alignment field to specify which way to align. Also, there should be a default that if there is no columns, rows or customalignmentLength, but alignment is set, then vertical means 1 column, horiziontal means 1 row and auto is... 1 column?

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
```diff
@@ -184,11 +184,10 @@ export class InputCheckbox<CustomSurvey, CustomHousehold, CustomHome, CustomPers
     };
 
     private getColumnedChoices = (choiceInputs: JSX.Element[]): JSX.Element[] => {
-        if (this.props.widgetConfig.columns === undefined || this.props.widgetConfig.columns <= 0) {
+        if ((this.props.widgetConfig.alignChoices === undefined || !this.props.widgetConfig.alignChoices)) {
```
tahini, line 187: Here, check columns, rows and customAlignmentLength instead, to drop alignChoices. This check is probably the cause of the test failure, as this widgetConfig is not set and the columns are thus not considered anymore

packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -0,0 +1,230 @@
...
+   * @returns A 2 dimension array containing the input array sorted into columns for display or the original array in case of invalid parameters.
+   */
+const sortByParameters = function (arr: any[], alignment?: 'vertical' | 'horizontal' | 'auto', columns?: number, rows?: number, customAlignmentLengths?: number[]): any[] | any[][] {
+    if (customAlignmentLengths && customAlignmentLengths.length > 0) {
+        switch (alignment) {
+            case 'vertical':
+                return customVertical(arr, customAlignmentLengths);
+            case 'horizontal':
+                return customHorizontal(arr, customAlignmentLengths);
+            default:
```
tahini, line 205: You should return a customVertical here

tahini (changes requested):
The new utility looks good, you just need to remove the alignChoices field from the widget config, make sure tests for InputRadio and Checkboc still run with the additions. And of course, run `yarn format` before submitting ;-)

Frost000:
For the default options I chose 2 because, that was what it was before this feature.

packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -0,0 +1,242 @@
+/*
+ * Copyright 2023, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { Type } from "typescript";
```
tahini, line 8: No need to import this

tahini (changes requested):
Change the generic type

packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -198,7 +196,7 @@ const customHorizontal = function<Type>(arr: Type[], custom: number[]): Type[][]
    * @param customAlignmentLengths - Specify the lengths of each row or column individually, if alignement is horizontal, the first specified row must be the longuest.
    * @returns A 2 dimension array containing the input array sorted into columns for display or the original array in case of invalid parameters.
    */
-const sortByParameters = function<Type>(arr: Type[], alignment?: 'vertical' | 'horizontal' | 'auto', columns?: number, rows?: number, customAlignmentLengths?: number[]): Type[] | Type[][] {
+const sortByParameters = function<T>(arr: T[], alignment?: 'vertical' | 'horizontal' | 'auto', columns?: number, rows?: number, customAlignmentLengths?: number[]): T[] | T[][] {
```
tahini, line 199: I think it is now impossible to have simply `T[]` as return type. The function now has a `else` with a `switch` that will always return one of the above method, so the return type should just be `T[][]`

packages/evolution-frontend/src/components/inputs/__tests__/InputChoiceSorting.test.ts
```diff
@@ -4,7 +4,9 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
+import exp from 'constants';
```
tahini, line 7: These imports look like automatic vscode auto-completed imports, I'm don't think they're used, same for 'ro' below

tahini (changes requested):
Remove unused imports and this should be good

tahini (approved):

PR #173, opened 2023-07-27
Questions: Add a parameter to join questions.

Frost000:
Fixes https://github.com/chairemobilite/evolution/issues/146
With the joinPrevious parameter, the previous question will be joined the one in which the option is stated.
joinedQuestions

### DIFF
--- a/example/demo_survey/src/survey/widgets/home.tsx
+++ b/example/demo_survey/src/survey/widgets/home.tsx
@@ -385,7 +385,6 @@ export const homeApartmentNumber = {
   path: "home.apartmentNumber",
   datatype: "string",
   twoColumns: true,
-  joinPrevious: true,
   label: {
     fr: '<p>Numéro d\'appartement<br /><span class="_em _pale">Laisser vide si aucun</span></p>',
     en: 'Apartment number<br /><span class="_em _pale">Leave empty if none</span>'
@@ -398,6 +397,7 @@ export const homeCity = {
   path: "home.city",
   twoColumns: true,
   datatype: "string",
+  joinWith: "homeRegion",
   label: {
     fr: "Ville",
     en: "City"
@@ -421,7 +421,7 @@ export const homeRegion = {
   path: "home.region",
   datatype: "string",
   twoColumns: true,
-  joinPrevious: true,
+  joinWith: "homeCountry",
   defaultValue: "Québec",
   label: {
     fr: "Province",
@@ -435,7 +435,6 @@ export const homeCountry = {
   path: "home.country",
   datatype: "string",
   twoColumns: true,
-  joinPrevious: true,
   defaultValue: "Canada",
   label: {
     fr: "Pays",
--- a/example/demo_survey/src/survey/widgets/home.tsx
+++ b/example/demo_survey/src/survey/widgets/home.tsx
@@ -385,6 +385,7 @@ export const homeApartmentNumber = {
   path: "home.apartmentNumber",
   datatype: "string",
   twoColumns: true,
+  joinPrevious: true,
   label: {
     fr: '<p>Numéro d\'appartement<br /><span class="_em _pale">Laisser vide si aucun</span></p>',
     en: 'Apartment number<br /><span class="_em _pale">Leave empty if none</span>'
@@ -420,6 +421,7 @@ export const homeRegion = {
   path: "home.region",
   datatype: "string",
   twoColumns: true,
+  joinPrevious: true,
   defaultValue: "Québec",
   label: {
     fr: "Province",
@@ -433,6 +435,7 @@ export const homeCountry = {
   path: "home.country",
   datatype: "string",
   twoColumns: true,
+  joinPrevious: true,
   defaultValue: "Canada",
   label: {
     fr: "Pays",
--- a/packages/evolution-frontend/src/components/survey/Question.tsx
+++ b/packages/evolution-frontend/src/components/survey/Question.tsx
@@ -326,7 +326,7 @@ export class Question<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> e
                 );
             }
         };
-        const content = (
+        const content = [(
             <div
                 style={{ position: 'relative' }}
                 className={`apptr__form-container${twoColumns ? ' two-columns' : ''}${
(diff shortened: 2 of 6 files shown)


example/demo_survey/src/survey/widgets/home.tsx
```diff
@@ -385,6 +385,7 @@ export const homeApartmentNumber = {
   path: "home.apartmentNumber",
   datatype: "string",
   twoColumns: true,
+  joinPrevious: true,
```
tahini, line 388: Interesting approach, but instead of joinPrevious, maybe use 'joinWith', which allows to specify which question to join it with. What if the previous one is actually invisible and it ends up being joined with the wrong other question?

But if so, that would complicate everything because it would need to be done at the Section level instead of Question

packages/evolution-frontend/src/components/survey/Question.tsx
```diff
@@ -392,7 +392,15 @@ export class Question<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> e
                     )}
                 </InputWidgetWrapper>
             </div>
-        );
+        )];
+
+        if ((widgetConfig as any).joinPrevious) {
```
tahini, line 397: Ok, so it is a visual hack to display the widgets together

example/demo_survey/src/survey/widgets/home.tsx
kaligrafy: I would suggest joinWith and if the joined with is not previous, just ignore

packages/evolution-frontend/src/components/survey/Question.tsx
kaligrafy: The best way would be to create nested arrays for the widgets in the section config instead of a simple array, but it may be more difficult to implement

example/demo_survey/src/survey/widgets/home.tsx
Frost000: I will try to see what I can do for join with. I made join with previous because questions will usually be hidden until an answers in that is before it is answered.

packages/evolution-legacy/src/components/survey/Section.js
```diff
@@ -166,6 +167,17 @@ export class Section extends React.Component {
         widgetStatus.errorMessage = widgetStatus.errorMessage || serverErrorMessage;
       }
 
+      const join = undefined;
```
tahini, line 170: Should be `let` instead of `const` if you reassign it later

packages/evolution-legacy/src/components/survey/Section.js
```diff
@@ -166,6 +167,17 @@ export class Section extends React.Component {
...
+      if (widgetConfig.joinWith) {
+        const previous = this.props.widgets[i - 1] ? this.props.widgets[i - 1] : undefined;
```
tahini, line 172: just `this.props.widgets[i-1]` It will be undefined anyway.

packages/evolution-legacy/src/components/survey/Section.js
```diff
@@ -166,6 +167,17 @@ export class Section extends React.Component {
...
+        const previousStatus = this.props.widgets[i - 1] ? _get(this.props.interview, `widgets.${this.props.widgets[i - 1]}`, {}) : undefined;
+        const next = this.props.widgets[i + 1] ? this.props.widgets[i + 1] : undefined;
+        const nextStatus = this.props.widgets[i + 1] ? _get(this.props.interview, `widgets.${this.props.widgets[i + 1]}`, {}) : undefined;
+
+        join = widgetConfig.joinWith === previous && previousStatus.isVisible ? 'previous' : undefined;
+        join = widgetConfig.joinWith === next && nextStatus.isVisible? 'next' : undefined;
```
tahini, line 178: You are reassigning the join on this line, so if it's the previous, it will never work. Use
`widgetConfig.joinWith === previous && previousStatus.isVisible ? 'previous' : widgetConfig.joinWith === next && nextStatus.isVisible? 'next' : undefined;`

packages/evolution-frontend/src/components/survey/Question.tsx
```diff
@@ -392,7 +394,28 @@ export class Question<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> e
...
             </div>
-        );
+        )];
+
+        switch(this.props.join) {
+            case "next":
+                content.push(
+                <div
+                    key="postJoin"
+                    style={{marginTop: '-40px', marginBottom: '-40px', paddingTop: '20px', paddingBottom: '20px'}}
```
tahini, line 404: next and previous are identical?

packages/evolution-frontend/src/components/survey/Question.tsx
```diff
@@ -326,8 +327,9 @@ export class Question<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> e
                 );
             }
         };
-        const content = (
+        const content = [(
             <div
+                key="content"
```
tahini, line 332: keys are used to uniquely identify a react object and allow to only refresh the ones that change. They should be unique in the application, you can this `this.props.path` as suffix to the key, for example `content_${this.props.path}`

tahini (changes requested):
excellent, it should work now. Just a few small changes remaining.

packages/evolution-frontend/src/components/survey/Question.tsx
Frost000: Yes, the html code pulls from the top and bottom question evenly to join them, so it is only a matter of putting it before or after the question.

tahini: oh ok! one is push the other unshift, ok. Maybe worth a comment in the code then.

packages/evolution-legacy/src/components/survey/Section.js
```diff
@@ -166,6 +167,16 @@ export class Section extends React.Component {
...
       }
 
+      let join = undefined;
+      if (widgetConfig.joinWith) {
+        const previous = this.props.widgets[i - 1];
+        const previousStatus = this.props.widgets[i - 1] ? _get(this.props.interview, `widgets.${this.props.widgets[i - 1]}`, {}) : undefined;
+        const next = this.props.widgets[i + 1];
+        const nextStatus = this.props.widgets[i + 1] ? _get(this.props.interview, `widgets.${this.props.widgets[i + 1]}`, {}) : undefined;
+
+        widgetConfig.joinWith === previous && previousStatus.isVisible ? 'previous' : widgetConfig.joinWith === next && nextStatus.isVisible? 'next' : undefined;
```
tahini, line 177: Misssing `join = ` at the beginning of the line

tahini (approved):

PR #174, opened 2023-07-28
Fix language not saved

davidmurray:
Based on my small scale testing, this seems to fix the issue. In my database, _language was always null, but now it seems to correctly show "fr" or "en". Fixes https://github.com/chairemobilite/od_mtl_2023/issues/374

@kaligrafy @tahini : any clue why the code was like that in the first place? Since the migration from transition-legacy, I can't git blame properly (or am I unaware of a feature of git?)

### DIFF
--- a/packages/evolution-frontend/src/actions/Survey.ts
+++ b/packages/evolution-frontend/src/actions/Survey.ts
@@ -24,6 +24,7 @@ import { prepareWidgets } from './utils';
 import { UserFrontendInterviewAttributes } from '../services/interviews/interview';
 import { incrementLoadingState, decrementLoadingState } from './LoadingState';
 import { CliUser } from 'chaire-lib-common/lib/services/user/userType';
+import i18n from 'chaire-lib-frontend/lib/config/i18n.config';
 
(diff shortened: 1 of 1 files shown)


kaligrafy:
No idea. It worked for some interviews though

kaligrafy:
I thinkg there was a bug with i18n that was not returning the language. Can you confirm it is correctly set by i18n on every launch?

tahini (changes requested):
I think a change in i18n configuration now allows to better detect the language at startup, which was not the case before. So it should work now.

You'll probably have to mock the i18n package in the failing test though.

kaligrafy:
@tahini why it does not compile?

tahini:
> @tahini why it does not compile?

It compiles, but the test fails because the i18n needs to be mocked.

kaligrafy:
Je te l'assigne @tahini, tu pourra sme montrer comment faire le mock pour la prochaine fois :-)

tahini:
Replaced with #212 which adds the test fix

PR #176, opened 2023-07-29
Add an evolution-interviewer node module for interviewer support

### LINKED ISSUE #175: Add an interviewer module in evolution
The Montreal and Quebec recent surveys both used phone interviewers for the survey.

We had to add specific roles and monitoring options for interviewers in both interviews. We could add a node module, which would be small at first, that can be included in the surveys that have phone interviewers. That would avoid duplicating code for interviewers and allow new features for one to be available in the other.

tahini:
Fixes #175

This interviewer module adds roles for interviewers, which are users allowed to read or update any interviews and interviewer supervisor, which have the same permissions as interviewers, but also manage interviewers.

This module also adds support for an interviewer mode in the interview, which can be used to change labels for this mode, or by widgets.

To use, the server admin application needs to call the `roleDefinition` default function to add the roles.

The frontend admin application needs to import the `interviewerSupport` file and add the interviewer options to the application configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
         "packages/evolution-backend",
         "packages/evolution-frontend",
         "packages/evolution-legacy",
+        "packages/evolution-interviewer",
         "example/demo_survey"
     ],
     "scripts": {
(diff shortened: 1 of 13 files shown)


tahini:
@kaligrafy I've put the obvious in this module so far. More will come, but it will avoid code duplication for everything concerning the interviewers, especially for the monitoring sections to come (for the interviewer supervisor roles, also available to admins)

kaligrafy (approved):

PR #186, opened 2023-08-03
Implements survey generator

Wanadian:


packages/evolution-common/src/services/surveyGeneration/SurveyGenerator.ts
```diff
@@ -0,0 +1,21 @@
+import { SurveyWidgetConfig } from "../../../../../example/demo_survey/src/config/widgets.config";
```
tahini, line 1: The function should rather be generic, it cannot depend on a file from the application itself

packages/evolution-common/src/services/surveyGeneration/CSVHandle.ts
```diff
@@ -0,0 +1,21 @@
+import fs = require('fs');
+import Papa = require('papaparse');
+import { SurveyWidgetConfig } from "../../../../../example/demo_survey/src/config/widgets.config";
```
tahini, line 3: We have a parseCsvFile in chaire-lib-common (or chaire-lib-backend for a server side version) which wraps papaparse for our internal API. Please use this instead

packages/evolution-common/src/services/surveyGeneration/SurveyGenerator.ts
```diff
@@ -0,0 +1,21 @@
...
+import { readCSVFile } from './CSVHandle';
+
+
+export async function WidgetListGenerator (filePath: string): Promise<SurveyWidgetConfig[]> {
+    const widgetList : SurveyWidgetConfig[] = [];
+    await readCSVFile(filePath)
+        .then((data) => {
+            data.length !== 0 && data.forEach((widgetProperties) => {
+                const widget : SurveyWidgetConfig = { ...widgetProperties as SurveyWidgetConfig };
```
tahini, line 10: There is no preparation or validation of the data, it's just the csv as is used for widget properties?

samuel-duhaime:
I think we can close, we now have a Generator to do this functionality, and we should focus on evolution-generator if we want to improve something.

PR #199, opened 2023-08-15
make sure loading state is 0 before focusing on invalid field (fix issue 161)

kaligrafy:
This needs TESTING because it could have SIDE EFFECTS!

### DIFF
--- a/packages/evolution-legacy/src/components/survey/Section.js
+++ b/packages/evolution-legacy/src/components/survey/Section.js
@@ -56,7 +56,7 @@ export class Section extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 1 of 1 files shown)


tahini:
What tests need to be done and what kind of side effects should tests look for?

kaligrafy:
We just need to make sure it always work when a question is invalid anytime in the workflow

tahini (changes requested):
It seems to indeed fix issue #161, but could you add to the commit message why it does so? Why does changing a response produce a loadingState > 0, but clicking the submit button brings the loadingState to 0? it's hard to see the link between the issue and the solution right now...

tahini (approved):

PR #221, opened 2023-08-25
add a way to keep the assigned value intact

kaligrafy:
when toggling between hide/show conditional status, this is useful if you want to use two different widget questions for the same value, and toggle between them according to a condition. Without this config, whenever the condition is switched back to show after a hide, the hide assigned value is replaced by the default or empty value. This is also useful to hide a question when a condition is met, but keep its assigned value active between hide and show toggles

### DIFF
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -319,6 +319,14 @@ export type QuestionWidgetConfig<CustomSurvey, CustomHousehold, CustomHome, Cust
     path: InterviewResponsePath<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     containsHtml?: boolean;
     label: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
+
+    /**
+    * When the conditional triggers a hide status and then a show status later on,
+    * instead of reverting to default or empty value, use the assigned value of the conditional:
+    * this is useful when we want to hide a widget but keep its assigned value intact after hide/show toggle.
+    */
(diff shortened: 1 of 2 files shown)


packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -319,6 +319,10 @@ export type QuestionWidgetConfig<CustomSurvey, CustomHousehold, CustomHome, Cust
     path: InterviewResponsePath<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     containsHtml?: boolean;
     label: I18nData<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
+    // When the conditional triggers a hide status and then a show status later on,
```
tahini, line 322: If you put this comment inside `/** blabla */`, then the text will be used as documentation for this field. The way you do now is a simple comment in the code. Since you went through the trouble of documenting this value, you might as well put it in `/** */`

tahini (changes requested):
looks simple enough, just update for documentation

packages/evolution-common/src/services/widgets/WidgetConfig.ts
kaligrafy: done

tahini (commented):
What's happening in this commit? Why all the refactoring? Did you run `yarn format`? Adding documentation for a field can't possibly trigger all those formatting changes? Or does it?

kaligrafy:
Mon VSC format automatiquement maintenant quand je sauvegarde... Je ne sais pas pourquoi...

tahini (approved):

PR #222, opened 2023-08-27
disable form submit default so it does not trigger on any key

kaligrafy:
we do not use the submit event at all, since buttons have their own onClick actions, and the enter/space key trigger on these button is correctly sent to onClick action for screen readers and ARIA

### DIFF
--- a/packages/evolution-legacy/src/components/survey/Survey.js
+++ b/packages/evolution-legacy/src/components/survey/Survey.js
@@ -173,7 +173,7 @@ export class Survey extends React.Component {
                   containsHtml       = {false}
                 />
               </div>)}
-              <form onSubmit={(e) => { if (e.key === 'enter' || e.which === 13 ) { e.preventDefault(); }}} className="apptr__form" id="survey_form" style={customStyle} autocomplete="off">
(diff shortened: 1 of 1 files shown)


packages/evolution-legacy/src/components/survey/Survey.js
```diff
@@ -173,7 +173,7 @@ export class Survey extends React.Component {
                   containsHtml       = {false}
                 />
               </div>)}
-              <form onSubmit={(e) => { if (e.key === 'enter' || e.which === 13 ) { e.preventDefault(); }}} className="apptr__form" id="survey_form" style={customStyle} autocomplete="off">
+              <form onSubmit={(e) => e.preventDefault() } className="apptr__form" id="survey_form" style={customStyle} autocomplete="off">
```
tahini, line 176: But the prior condition should have been enough to prevent the enter key from submitting the form, no? Why did it submit anyway?

tahini (approved):
It works, thanks

packages/evolution-legacy/src/components/survey/Survey.js
kaligrafy: it seems the synthetic event triggered by enter key in an input field does not propagate as a key event... The key presed is lost.

PR #224, opened 2023-08-28
interviewer: Let the apps add the `interviewer` path app.use

tahini:
Otherwise, it clashes with the `/` route for some reason and we get the main application page.


tahini:
@kaligrafy, this is required for Quebec admin site

kaligrafy (approved):

PR #228, opened 2023-08-29
InputMapPoint/InputMapFindPlace: add parameter to hide/show geocoding…

kaligrafy:
… button


kaligrafy:
Forget this PR, if we set the geocoding label to null, it will not be shown anyway.

kaligrafy:
Use #229 instead

PR #229, opened 2023-08-29
InputMapPoint/InputMapFindPlace: add parameter to hide/show geocoding…

kaligrafy:
… button

### DIFF
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -284,6 +284,7 @@ export type InputMapPointType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     CustomPerson
 > & {
     inputType: 'mapPoint';
+    showSearchPlaceButton?: boolean | ParsingFunction<boolean, CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
 };
 
 export type InputMapFindPlaceType<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> = InputMapType<
@@ -293,6 +294,7 @@ export type InputMapFindPlaceType<CustomSurvey, CustomHousehold, CustomHome, Cus
     CustomPerson
 > & {
     inputType: 'mapFindPlace';
+    showSearchPlaceButton?: boolean | ParsingFunction<boolean, CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
     placesIcon?: {
         url: string | ParsingFunction<string, CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
         size?: [number, number];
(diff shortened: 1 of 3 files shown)


packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -284,6 +284,7 @@ export type InputMapPointType<CustomSurvey, CustomHousehold, CustomHome, CustomP
     CustomPerson
 > & {
     inputType: 'mapPoint';
+    showSearchPlaceButton?: boolean | ParsingFunction<string, CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
```
tahini, line 287: It should be `ParsingFunction<boolean>`, not string

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
```diff
@@ -38,7 +38,7 @@ export type InputMapFindPlaceProps<CustomSurvey, CustomHousehold, CustomHome, Cu
     widgetConfig: InputMapFindPlaceType<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>;
 };
 
-interface InputMapFindPlaceState {
+interface InputMapFindPlaceState<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> {
```
tahini, line 41: Does this interface need to be generic? The generic types don't appear to be used

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
```diff
@@ -409,7 +409,7 @@ export class InputMapFindPlace<CustomSurvey, CustomHousehold, CustomHome, Custom
                     readOnly={true}
                     ref={this.props.inputRef}
                 />
-                {this.props.widgetConfig.refreshGeocodingLabel && (
+                {this.props.widgetConfig.refreshGeocodingLabel && this.props.widgetConfig.showSearchPlaceButton !== false && (
```
tahini, line 412: showSearchPlaceButton can be a ParsingFunction, I'd expect to see a call to `surveyHelper.parseBoolean` somewhere

tahini (changes requested):

packages/evolution-frontend/src/components/inputs/InputMapFindPlace.tsx
tahini: > Does this interface need to be generic? The generic types don't appear to be used

The question remains here

tahini (approved):
Ok, it's just the matter of the generic state type, but otherwise, looks good

kaligrafy:
Il m'a forcé à ajouter les generic types au state, il voulait pas compiler sans ça... Je ne sais pas à quoi ça sert...

PR #237, opened 2023-09-06
allow keyboard to trigger search place button, closes #236

### LINKED ISSUE #236: mapFindPlace: Search button not clickable with keyboard
mapPoint works fine though

But selecting the button and clicking space or enter should trigger the click action

kaligrafy:


tahini:
It works to click on the button, but after it comes back with choices, clicking  on the tab key to navigate to the next element doesn't work, it's like we're not focused anywhere anymore.

kaligrafy:
That is a completely separate bug in my opinion!

tahini (approved):
I don't know if it's a separate issue, because it works fine with MapPoint, but not MapFindPlace, unless I'm completely mistaken. But we can fix it separately if you prefer

kaligrafy:
Can you create the second issue, becuase I am not sure how to explain it correctly...

PR #255, opened 2023-09-15
Fix 156: style InputRadioNumber better

davidmurray:


tahini (changes requested):
Looks good, but the test failure is legit, you may need to update the snapshots.

Also, could you add a screenshot of the new look?

davidmurray:
@tahini 

image

Fixed tests!

tahini (approved):

PR #259, opened 2023-09-19
Fix 156: add disallowedCharactersRegex feature

davidmurray:


packages/evolution-frontend/src/components/inputs/InputString.tsx
```diff
@@ -47,12 +47,14 @@ export const InputString = <CustomSurvey, CustomHousehold, CustomHome, CustomPer
             style={{ textTransform: props.widgetConfig.textTransform || ('none' as any) }}
             autoComplete="none"
             type="text"
+            pattern={props.widgetConfig.numericKeyboard ? "[0-9]*" : ""}
```
tahini, line 50: How does this one work? I still can enter non numeric characters in  the car number input.

tahini (commented):
Looks good, but the numericKeyboard does not work for the car number. I'm not sure what it's supposed to do exactly.

davidmurray:
@tahini it's two different things:
* `pattern = [0-9]*` allows bringing up the numeric keyboard on mobile devices. It also has something to do with HTML form validation, but we are not using that anyway.
* `disallowedCharactersRegex` is for restricting entry to numbers only.

If you just use `pattern`, then on mobile you get the numeric keyboard but you can still enter non-digit characters on desktop. If you just use `disallowedCharactersRegex` with an appropriate regex, you can only enter digits, but you get the full keyboard on mobile devices.
Ideally you combine both options for a good UX

tahini (approved):
Thanks for the precision

PR #265, opened 2023-09-21
Use latest lodash instead of per-method packages

tahini:


tahini:
For the record, this approach slightly decreases memory usage of the bundle (4.19 MB for the gzipped dev build instead of 4.37), so no memory increase here

PR #281, opened 2023-09-29
add parser function and path config

kaligrafy:
for better access to configuration of each path, which will be useful in the future when we will merge all widget/path configs for both the questionnaire and the admin with validations

### DIFF
--- a/packages/evolution-common/src/services/parsers/ParserFunction.ts
+++ b/packages/evolution-common/src/services/parsers/ParserFunction.ts
@@ -1,0 +1,28 @@
+/*
+ * Copyright 2023, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { UserInterviewAttributes, InterviewResponses, Household, Person } from '../interviews/interview';
+
+/**
+ * Parser function to change/manipulate/enhance values for export.
+ * Associated with a specific path.
+ * TODO: update this parser made for legacy purpose, one step at a time,
+ * using new classes for person, household, etc. and after validations
+ * cleaning.
+ */
+export type ParserFunction<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> = (
+    interview: UserInterviewAttributes<CustomSurvey, CustomHousehold, CustomHome, CustomPerson>,
+    responses: InterviewResponses<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> & { [key: string]: any },
+    household: Household<CustomHousehold, CustomPerson>,
+    home: {
+        region?: string;
+        country?: string;
+    } & CustomHome,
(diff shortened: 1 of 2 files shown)


packages/evolution-common/src/services/parsers/ParserFunction.ts
```diff
@@ -0,0 +1,14 @@
+import { UserInterviewAttributes, InterviewResponses, Household, Person } from '../interviews/interview';
+
+export type ParserFunction<CustomSurvey, CustomHousehold, CustomHome, CustomPerson> = (
```
tahini, line 3: Please add a comment what will this function be used for. It receives in parameter the personsById AND the personsArray, which are redundant, and a single Person, and a Household, which should contain the personsById directly, so I'm not sure all these parameters are required

packages/evolution-common/src/services/widgets/PathConfig.ts
```diff
@@ -0,0 +1,32 @@
+import { ValidationFunction } from './WidgetConfig';
+import { ParserFunction } from '../parsers/ParserFunction';
+
+/**
+ * Path config include custom admin validations,
+ * flag to include the path in export files or not,
+ * and parsers for custom changes to data before export
+ * Keys are paths with uuids replaced by '@' and array indexes
```
tahini, line 8: There is no keys in this type, maybe add another type that describe the object where you will use the PathConfig type, with the info for the keys there

packages/evolution-common/src/services/widgets/PathConfig.ts
```diff
@@ -0,0 +1,32 @@
+import { ValidationFunction } from './WidgetConfig';
+import { ParserFunction } from '../parsers/ParserFunction';
+
+/**
+ * Path config include custom admin validations,
```
tahini, line 5: This does not explain the why of this type:

"This type describes the paths in the responses field, with data for validation and export". Keep the rest of the description after.

tahini (changes requested):
Also, add the copyright headers to the files.

There's a bit of documentation missing or not quite right.

packages/evolution-common/src/services/widgets/PathConfig.ts
kaligrafy: added

packages/evolution-common/src/services/parsers/ParserFunction.ts
```diff
@@ -0,0 +1,25 @@
...
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { UserInterviewAttributes, InterviewResponses, Household, Person } from '../interviews/interview';
+
+/**
+ * Parser function to change/manipulate/enhance values for export.
+ * Associated with a specific path.
```
tahini, line 12: Manque juste le TODO pour expliquer que cette fonction type une utilisation legacy, mais devrait être mise à jour au fur et à mesure qu'on nettoie les validations...

tahini (approved):

PR #286, opened 2023-10-03
Create base classes for surveyed objects

kaligrafy:


packages/evolution-common/src/services/baseObjects/Surveyable.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export type SurveyableAttributes = {
+
+    sample: Sample;
+    survey: Survey;
+    sampleBatchNumber?: number;
+
+}
+
+export class Surveyable extends Uuidable {
```
tahini, line 20: What is this? Who would be surveyable?

packages/evolution-common/package.json
```diff
@@ -24,7 +24,8 @@
     "chaire-lib-common": "^0.2.2",
     "i18next": "^22.4.15",
     "lodash": "^4.17.21",
-    "typescript": "^4.9.4"
+    "typescript": "^4.9.4",
+    "uuid": "^9.0.1"
```
tahini, line 28: This adds a dependency to yet another uuid package. Please use the same as other packages, or update all of them everywhere (including transition, you might prefer just using the same as elsewhere)

tahini (changes requested):
Please run `yarn format` There seems to be a bit of wrong formatting. Also, that's a lot of types... that don't apply to all surveys. But keep going, it looks like a completely new module! That will be used for the city simulator later on!

packages/evolution-common/src/services/baseObjects/BasePerson.ts
```diff
@@ -0,0 +1,167 @@
...
+import { Tripable } from './Tripable';
+import { Uuidable } from './Uuidable';
+import { BasePlace } from './BasePlace';
+import { BaseVisitedPlace } from './BaseVisitedPlace';
+import { BaseTrip } from './BaseTrip';
+import { BaseVehicle } from './BaseVehicle';
+import * as PAttr from './attributeTypes/PersonAttributes';
+import { Vehicleable } from './Vehicleable';
+
+type BasePersonAttributes = {
```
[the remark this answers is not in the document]
kaligrafy: I took what is in my thesis, so this is like "official"???!!!

packages/evolution-common/src/services/baseObjects/BasePlace.ts
```diff
@@ -0,0 +1,109 @@
...
+    _isValid: OptionalValidity;
+
+    geography: GeoJSON.Feature<GeoJSON.Point> | undefined;
+    name?: string;
+    shortname?: string;
+    address?: BaseAddress;
+    osmId?: string; // OpenStreetMap id (can be a node, way or relation). This is the short version: n1234, w2345, r3456
+    landRoleId?: string; // official land role building/location id
+    postalId?: string; // official postal id
+    buildingId?: string; // official building id
```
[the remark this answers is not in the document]
kaligrafy: THe official id of the data in the country official database.
Most postal service have specific ids for each address. Same for landrole and the building id, it depends of the country, but some have a specific id for each building.

packages/evolution-common/src/services/baseObjects/Surveyable.ts
kaligrafy: Added a comment

packages/evolution-common/package.json
kaligrafy: You mean you want me to update all uuid versions in other packages to 9.0.1? Wouldn't it be better to wait for current surveys to be completed?

packages/evolution-common/src/services/baseObjects/Surveyable.ts
kaligrafy: BaseINterview is surveyable

kaligrafy:
I added attributes to trip chain (I forgot them)

packages/evolution-common/package.json
kaligrafy: I normalized all uuid versions to 8.3.2

tahini (approved):

PR #300, opened 2023-10-10
convert create factory fucntion to return Error[]

kaligrafy:
+ add validateParams static function to validate all params sent as input to base objects

### DIFF
--- a/packages/evolution-common/src/services/baseObjects/BaseInterview.ts
+++ b/packages/evolution-common/src/services/baseObjects/BaseInterview.ts
@@ -7,6 +7,7 @@
 
 import { OptionalValidity, IValidatable } from './Validatable';
 import { Surveyable, SurveyableAttributes } from './Surveyable';
+import { Uuidable } from './Uuidable';
 import { BasePerson } from './BasePerson';
 import { BaseHousehold } from './BaseHousehold';
 import { BaseOrganization } from './BaseOrganization';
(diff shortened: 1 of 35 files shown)


packages/evolution-common/src/services/baseObjects/BaseTrip.ts
```diff
@@ -37,22 +37,22 @@ export type ExtendedTripAttributes = BaseTripAttributes & {[key: string]: any};
 export class BaseTrip extends Uuidable implements IValidatable {
 
     _isValid: OptionalValidity;
-    _weight?: Weight;
+    _weights?: Weight[];
```
tahini, line 40: What do weights represent? Why an array?

packages/evolution-common/src/services/baseObjects/BaseTrip.ts
```diff
@@ -70,20 +70,62 @@ export class BaseTrip extends Uuidable implements IValidatable {
      * Factory that validates input from an interview and makes
      * sure types and required fields are valid before returning a new object
      * @param dirtyParams
-     * @returns BaseTrip
+     * @returns BaseTrip | Error[]
```
tahini, line 73: Is `Error[]` enough? What will you do with it in the end? Just dump it somewhere for the analyst to see? Or would you need some more specific error type, with object class, object uuid and error message?

packages/evolution-common/src/services/baseObjects/BaseInterview.ts
```diff
@@ -69,21 +70,73 @@ export class BaseInterview extends Surveyable implements IValidatable {
...
+            errors.push(new Error('BaseInterview validateParams: invalid completedDate'));
+        }
+
+        // Validate completedTime (if provided)
+        if (dirtyParams.completedTime !== undefined && (typeof dirtyParams.completedTime !== 'number' || dirtyParams.completedTime < 0)) {
+            errors.push(new Error('BaseInterview validateParams: completedTime should be a non-negative number'));
+        }
+
+        // Validate baseHousehold (if provided)
+        if (dirtyParams.baseHousehold !== undefined && !(dirtyParams.baseHousehold instanceof BaseHousehold)) {
```
tahini, line 121: Shouldn't the interview be responsible to construct its own household? Now the dirty params seems to come with the household, which comes with the persons and trips and segments, etc. That means you have to dig down the object hierarchy to create to deepest objects first. You might want to let the interview create the household, and the household create the persons from the dirty params. The factory function would create the contained objects first

tahini (commented):
Some questions and thoughts

packages/evolution-common/src/services/baseObjects/BaseInterview.ts
kaligrafy: Weighting a sample over the reference population is the usual process involved when dealing with statistically based surveys. We need to make sure that the results are representative of the overall population associated with the survey. Each main object of survey data can be weighted, and each can include distinct weighting methods, so this is why we need to accept an array of weights, which includes the weight itself (a decimal number) and the weighting method. I am not sure we should include this in the documentation for now, since this is pretty usual stuff. All Origin-Destination and travel surveys are weighted.

packages/evolution-common/src/services/baseObjects/BaseTrip.ts
kaligrafy: Weighting a sample over the reference population is the usual process involved when dealing with statistically based surveys. We need to make sure that the results are representative of the overall population associated with the survey. Each main object of survey data can be weighted, and each can include distinct weighting methods, so this is why we need to accept an array of weights, which includes the weight itself (a decimal number) and the weighting method. I am not sure we should include this in the documentation for now, since this is pretty usual stuff. All Origin-Destination and travel surveys are weighted.

kaligrafy:
I added TODOs for Error specialization

tahini (approved):

PR #313, opened 2023-10-12
disable add grouped object when loading state is not empty

kaligrafy:


tahini (commented):
Still remains to be tested if it fixes the problem or not.

tahini (approved):
I manually added a delay after the loading state increment in the updateInterview redux action and I could reproduce the problem and this fixed it.

PR #330, opened 2023-10-20
Generic widgets

davidmurray:
@tahini  @kaligrafy  please suggest improvements ! 🙂

### DIFF
--- a/packages/evolution-frontend/src/components/genericWidgets/home.tsx
+++ b/packages/evolution-frontend/src/components/genericWidgets/home.tsx
@@ -1,0 +1,61 @@
+import { TFunction } from 'i18next';
+import { getI18nContext } from 'evolution-interviewer/lib/client/config/i18nextExtra.config';
+import { _isBlank, _isEmail } from 'chaire-lib-common/lib/utils/LodashExtensions';
+
+export const accessCode: any = {
+    type: 'question',
+    twoColumns: true,
+    path: 'accessCode',
+    inputType: 'string',
+    datatype: 'string',
+    containsHtml: true,
+    inputFilter: (input: string) => {
+        input = input.replace("_", "-"); // change _ to -
+        input = input.replace(/[^-\d]/g, ''); // Remove everything but numbers and -
+        // Get only the digits. If we have 8, we can automatically format the access code.
+        const digits = input.replace(/\D+/g, '');
+        if (digits.length === 8) {
+            return digits.slice(0, 4) + "-" + digits.slice(4);
+        }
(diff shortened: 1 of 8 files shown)


kaligrafy (approved):

packages/evolution-frontend/src/components/genericWidgets/home.tsx
```diff
@@ -0,0 +1,61 @@
...
+    },
+    label: (t: TFunction) => t('genericWidgets:home:ContactEmail', { context: getI18nContext() }),
+    validations: function (value) {
+        return [
+            contactEmailInvalidAddressValidation(value),
+        ];
+    }
+};
+
+export const contactEmailInvalidAddressValidation = function(value: string) {
```
tahini, line 56: Maybe don't include "contact" in the name here, as it could be used for other email widgets. `isEmailValidValidation` maybe?

packages/evolution-frontend/src/components/genericWidgets/home.tsx
```diff
@@ -0,0 +1,61 @@
...
+            }
+        ];
+    }
+};
+
+export const contactEmail = {
+    type: 'question',
+    path: 'contactEmail',
+    inputType: 'string',
+    datatype: 'string',
```
tahini, line 39: You can use your new `keyboardInputMode` here too

tahini (changes requested):
Small details to fix, otherwise, looks good.

@samuel-duhaime Do you think this approach is compatible with the data-driven surveys you are working on? We could have in evolution generic configurable high-level widgets that can be [metaphorically] dragged-and-dropped in the surveys, to avoid having to define them completely and have "nice defaults" (for labels, formats, types, paths, etc).

samuel-duhaime:
> Small details to fix, otherwise, looks good.
> 
> @samuel-duhaime Do you think this approach is compatible with the data-driven surveys you are working on? We could have in evolution generic configurable high-level widgets that can be [metaphorically] dragged-and-dropped in the surveys, to avoid having to define them completely and have "nice defaults" (for labels, formats, types, paths, etc).

Yes @tahini @davidmurray , it's very easy to implement that. In the Excel widgets tab, we just need to have an InputType "Generic", and we are good to go. :)

greenscientist:
@tahini , @samuel-duhaime what is the status here?

tahini:
Anyway, with the widgets moving to evolution, this will change

tahini:
Closing, as current work makes this deprecated and it would just be harder to rebase anyway

PR #332, opened 2023-10-20
Add an audit table, queries and utility functions

tahini:


packages/evolution-backend/src/models/__tests__/audits.db.test.ts
```diff
@@ -0,0 +1,270 @@
...
+        objectType: 'person',
+        objectUuid: person1Uuid,
+        _version: 2,
+        errorCode: 'NewPersonErrorCode1',
+        message: 'NewMessage1',
+        ignore: false
+    }, {
+        objectType: 'person',
+        objectUuid: person2Uuid,
+        _version: 2,
```
kaligrafy, line 132: I think we should remove the underscore... not sure why I put that there!

packages/evolution-backend/src/models/migrations/20231019213400_createAuditTbl.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const tableName = 'sv_audits';
+
+export async function up(knex: Knex): Promise<unknown> {
+    await knex.schema.createTable(tableName, (table: Knex.TableBuilder) => {
+        table.integer('interview_id').notNullable().index();
+        table.foreign('interview_id').references('sv_interviews.id').onDelete('CASCADE');
+        table.string('error_code').notNullable();
+        table.string('object_type').notNullable();
+        // uuid cannot be nullable if we want it as primary key. Code will fallback to using the interview's uuid for single objects so that it remains unique per interview/object_type
```
kaligrafy, line 18: It's ok, because by definition, when there is no object uuid, it means the audit is on the interview itself.

kaligrafy (approved):

PR #333, opened 2023-10-21
InputChoiceSorting: Rework logic to work horizontally.

Frost000:
Rework logic to use better css classes and work with Flex to be able to set custom lengths freely horizontally. The horizontal limitation of the older logic is no more and should be more flexible.

Ex:
CustomSortSansLimites

### DIFF
--- a/packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
+++ b/packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
@@ -203,11 +203,19 @@ export class InputCheckbox<CustomSurvey, CustomHousehold, CustomHome, CustomPers
         const columnedChoiceInputs: JSX.Element[] = [];
         for (let i = 0, count = widgetsByColumn.length; i < count; i++) {
             const columnWidgets = widgetsByColumn[i];
-            columnedChoiceInputs.push(
-                <div className="survey-question__input-checkbox-group-column" key={i}>
-                    {columnWidgets}
-                </div>
-            );
+            if (this.props.widgetConfig.alignment === undefined || this.props.widgetConfig.alignment === 'vertical') {
+                columnedChoiceInputs.push(
+                    <div className="survey-question__input-checkbox-group-column" key={i}>
+                        {columnWidgets}
+                    </div>
+                );
+            } else {
+                columnedChoiceInputs.push(
+                    <div className="survey-question__input-checkbox-group-row" key={i}>
+                        {columnWidgets}
+                    </div>
+                );
+            }
         }
 
         return columnedChoiceInputs;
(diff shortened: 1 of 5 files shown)


tahini (changes requested):
Thanks! seems to work, I,m no css expert, maybe @kaligrafy or @davidmurray can take a look. A few comments on the code though, and please run `yarn format` before submitting.

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
```diff
@@ -263,6 +271,47 @@ export class InputCheckbox<CustomSurvey, CustomHousehold, CustomHome, CustomPers
...
+                {this.props.customId && (
+                    <div className="label-input-custom-container">
+                        {strCustomLabel && (
+                            <label htmlFor={this.props.customId}>
+                                <span>{strCustomLabel}</span>
+                            </label>
+                        )}
+                        <input
+                            type="text"
+                            tabIndex={-1}
```
davidmurray, line 300: peut-être que ce code va faire en sorte que l'élément de sera pas sélectionnable avec le clavier. Est-ce le cas?
Si oui il faudrait retirer `tabIndex={-1}`

packages/evolution-frontend/src/components/inputs/InputChoiceSorting.ts
```diff
@@ -200,41 +98,27 @@ const sortByParameters = function <Value>(
...
     } else {
-        switch (alignment) {
-        case 'vertical':
-            return verticalByColumns(arr, 2);
-        case 'horizontal':
-            return horizontalByRows(arr, 2);
-        default:
-            return verticalByColumns(arr, 2);
-        }
+        return splitSort(arr, 2);
```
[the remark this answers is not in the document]
Frost000: Yes

packages/evolution-frontend/src/components/inputs/__tests__/InputChoiceSorting.test.ts
```diff
@@ -4,87 +4,60 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
+import { DESTRUCTION } from 'dns';
```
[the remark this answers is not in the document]
tahini: Please remove

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
```diff
@@ -203,11 +203,19 @@ export class InputCheckbox<CustomSurvey, CustomHousehold, CustomHome, CustomPers
         const columnedChoiceInputs: JSX.Element[] = [];
         for (let i = 0, count = widgetsByColumn.length; i < count; i++) {
             const columnWidgets = widgetsByColumn[i];
-            columnedChoiceInputs.push(
-                <div className="survey-question__input-checkbox-group-column" key={i}>
-                    {columnWidgets}
-                </div>
-            );
+            if (this.props.widgetConfig.alignment === undefined || this.props.widgetConfig.alignment === 'vertical') {
```
tahini, line 206: Here, in this if, the undefined alignment is considered as vertical

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
```diff
@@ -273,11 +281,15 @@ export class InputCheckbox<CustomSurvey, CustomHousehold, CustomHome, CustomPers
             )
             : null;
 
+        const shouldDisplayAsRows =
+            this.props.widgetConfig.alignment === undefined || this.props.widgetConfig.alignment === 'auto';
```
tahini, line 285: In this one, the undefined should be displayed as rows.

1- Is it normal or a mistake? If normal, please add a comment explaining why in the code
2- The alignment type says `alignment?: 'vertical' | 'horizontal' | 'auto'`. I think you should add a `||` for the horizontal case here.

tahini (changes requested):
You may want to add a few unit tests for either InputRadio or InputCheckbox to make sure the right css is picked: in a `describe` block, add 4 tests with each of the alignments types, mixed with custom alignement length and columns/rows, etc. It should be fairly easy to add snapshot tests and make sure the result is as expected.

greenscientist:
@tahini , should we incorporate this or forget about it?

PR #335, opened 2023-10-24
audits: Add server config for applications to advertise their audits

tahini:
Survey can set an `interviewAudits` function that takes a set of responses and return an array of AuditForObject, using `setProjectConfig({ interivuewAudits: myFunc })`.

Audits are run on an interview when it is requested for validation, after copying the responses to validated_data. If an audit function is specified, it runs the audit on the validated_data, then save the responses to the database and returns them to the caller.

Currently, audits are sent back to the client with `/survey/validateInterview/:interviewUuid` API call

Also add an API call to update audits from the client, for example, to set the ignore status.


tahini:
@kaligrafy here's the poutine for the interview auditing. You just need to implement the auditInterview function in the survey and return the audit objects.

kaligrafy (approved):

PR #340, opened 2023-10-25
move convertParamsErrorsToAudits to common

kaligrafy:
this function is needed in frontend to show errors in the validation one pager


tahini:
In theory, the frontend should not do the audit, it already has everything that is needed in the auditsV2 parameter received from server. It should not be done twice.

kaligrafy:
Well, then we need to be able to validate the params type also, which is not the case right now.

kaligrafy:
Ignore this, these will be converted on backend

PR #341, opened 2023-10-26
add the missing create function to BasePerson

kaligrafy:
for some reason it was missing only in this class


packages/evolution-common/src/services/baseObjects/BasePerson.ts
```diff
@@ -136,6 +136,18 @@ class BasePerson extends Uuidable implements IBasePersonAttributes, IValidatable
         this.baseHome = params.baseHome;
     }
 
+
+    /**
+     * Factory that validates input from an interview and makes
+     * sure types and required fields are valid before returning a new object
+     * @param dirtyParams
+     * @returns BaseJourney | Error[]
```
tahini, line 144: returns BasePerson

tahini (changes requested):
small fix in the comments

tahini (approved):

PR #354, opened 2023-11-29
feat: add generateInputRange script for Generator

samuel-duhaime:
With this script, it's going to help to build InputRange slider. 

This is the documentation about InputRange: 

## Generate InputRange

In Excel, the `InputRange` tab generates sliders in `inputRange.tsx`. The provided table example will generate the following TypeScript code, defining the `confiantInputRange` object in `inputRange.tsx`.

### InputRange Fields

| Field          | Description                                     | Example                 | Type        |
|----------------|-------------------------------------------------|-------------------------|-------------|
| inputRangeName | Name of the input range                         | confiantInputRange      | string      |
| labelFrMin     | French label for the minimum value              | Pas du tout confiant    | string      |
| labelFrMax     | French label for the maximum value              | Très confiant           | string      |
| labelEnMin     | English label for the minimum value             | Not at all confident    | string      |
| labelEnMax     | English label for the maximum value             | Very confident          | string      |
| minValue       | Minimum numerical value for the input range     | -10                     | number      |
| maxValue       | Maximum numerical value for the input range     | 100                     | number      |
| unitFr         | Unit in French                                  | %                       | string      |
| unitEn         | Unit in English                                 | %                       | string      |

### InputRange TypeScript

```typescript
// inputRange.tsx
export const confiantInputRange = {
    labels: [
        {
            fr: 'Pas du tout confiant',
            en: 'Not at all confident'
        },
        {
            fr: 'Très confiant',
            en: 'Very confident'
        }
    ],
    minValue: -10,
    maxValue: 100,
    formatLabel: (value, language) => {
        return value + ' ' + (language === 'fr' ? '%' : '%');
    }
};
```

### DIFF
--- a/generator/scripts/generateInputRange.py
+++ b/generator/scripts/generateInputRange.py
@@ -1,0 +1,97 @@
+import openpyxl  # Read data from Excel
+
+
+# Function to replace single quotes and stringify text
+def replaces_quotes_and_stringify(text):
+    if text is not None:
+        return str(text).replace("'", "\\'")  # Replace single quotes
+    return None
+
+
+# Function to generate inputRange.tsx
+def generate_input_range(input_file: str, output_file: str):
+    try:
+        # Load Excel file
+        workbook = openpyxl.load_workbook(input_file, data_only=True)
+        sheet = workbook["InputRange"]  # Get InputRange sheet
+
+        # Get headers from the first row
+        headers = [cell.value for cell in list(sheet.rows)[0]]
+
+        # Generate TypeScript code
+        ts_code: str = ""  # TypeScript code to be written to file
+        indentation: str = "    "  # 4-space indentation
+
+        # Add imports
+        # ts_code += "import { Labels } from '../types/inputTypes';\n\n"
+
+        # Iterate through each row in the sheet, starting from the second row
+        for row in list(sheet.rows)[1:]:
+            # Create a dictionary from the row values and headers
+            row_dict = dict(zip(headers, (cell.value for cell in row)))
+
+            # Get values from the row dictionary
+            input_range_name = row_dict["inputRangeName"]
+            label_fr_min = replaces_quotes_and_stringify(row_dict["labelFrMin"])
+            label_fr_max = replaces_quotes_and_stringify(row_dict["labelFrMax"])
+            label_en_min = replaces_quotes_and_stringify(row_dict["labelEnMin"])
+            label_en_max = replaces_quotes_and_stringify(row_dict["labelEnMax"])
+            min_value = str(row_dict["minValue"])
+            max_value = str(row_dict["maxValue"])
+            unit_fr = replaces_quotes_and_stringify(row_dict["unitFr"])
+            unit_en = replaces_quotes_and_stringify(row_dict["unitEn"])
+
+            # Generate TypeScript code
+            ts_code += "export const " + input_range_name + " = {\n"
+            ts_code += indentation + "labels: [\n"
+            ts_code += indentation + indentation + "{\n"
+            ts_code += (
+                indentation
+                + indentation
+                + indentation
+                + "fr: '"
+                + label_fr_min
+                + "',\n"
+            )
+            ts_code += (
+                indentation + indentation + indentation + "en: '" + label_en_min + "'\n"
+            )
+            ts_code += indentation + indentation + "},\n"
+            ts_code += indentation + indentation + "{\n"
+            ts_code += (
+                indentation
+                + indentation
+                + indentation
+                + "fr: '"
+                + label_fr_max
+                + "',\n"
+            )
+            ts_code += (
+                indentation + indentation + indentation + "en: '" + label_en_max + "'\n"
+            )
+            ts_code += indentation + indentation + "}\n"
+            ts_code += indentation + "],\n"
+            ts_code += indentation + "minValue: " + min_value + ",\n"
+            ts_code += indentation + "maxValue: " + max_value + ",\n"
+            ts_code += indentation + "formatLabel: (value, language) => {\n"
+            ts_code += (
+                indentation
+                + indentation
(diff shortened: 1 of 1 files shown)


generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+
+# Function to replace single quotes and stringify text
+def replaces_quotes_and_stringify(text):
+    if text is not None:
+        return str(text).replace("'", "\\'")  # Replace single quotes
+    return None
+
+
+# Function to generate inputRange.tsx
+def generate_input_range(input_file: str, output_file: str):
```
tahini, line 12: Here you can split this function into a few parts:

1- Load the file, which is now an excel but can eventually be a csv or anything else
2- data sanitization (the part under your comment #Getvalues from the row dictionary)
3- The typescript generator itself, which takes the data and generates the code

If you look at the fillLocales.py script in Evolution, which has separate classes for reading input file and writing output data. Small classes and/or functions with clear responsibilities are to be favored

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+        # Add imports
+        # ts_code += "import { Labels } from '../types/inputTypes';\n\n"
+
+        # Iterate through each row in the sheet, starting from the second row
+        for row in list(sheet.rows)[1:]:
+            # Create a dictionary from the row values and headers
+            row_dict = dict(zip(headers, (cell.value for cell in row)))
+
+            # Get values from the row dictionary
+            input_range_name = row_dict["inputRangeName"]
```
tahini, line 34: Extra sanitation is probably required here before sending to typescript. The Evolution script should make sure the data entered is good. The one you did is, but to bring those scripts upstream, you should assume the user may make mistakes and you should try to catch as much as possible with nice error messages to help him/her improve the input data. 

What if there are spaces in the name ? You can use some slugify function (there's probably a lib for that in python) to sanitize the input.

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+            label_fr_min = replaces_quotes_and_stringify(row_dict["labelFrMin"])
```
tahini, line 35: Add a TODO to support more languages. Ideally, to prepare for this future support, rename to labelMin_fr/labelMin_en, such that the string after the `_` character is the locale name.

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+            ts_code += (
+                indentation
+                + indentation
+                + indentation
+                + "fr: '"
+                + label_fr_min
+                + "',\n"
+            )
+            ts_code += (
+                indentation + indentation + indentation + "en: '" + label_en_min + "'\n"
```
tahini, line 57: To avoid things like `indentation + indentation +...`, maybe just add a function `indent(num)`, which will return the right number of indentatation, so here you would read `indent(3) + "en: '" + label_en_min...`

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+            )
+            ts_code += indentation + indentation + "},\n"
```
tahini, line 59: Extract the `\n` to a `NEWLINE` constant. You'll be happier if a day you need to change that to the OS's newline character instead of \n, if ever...

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
+import openpyxl  # Read data from Excel
```
tahini, line 1: Add some doc on top to explain how this is meant to be used. Is it a library to include in the generator script?

generator/scripts/generateInputRange.py
```diff
@@ -0,0 +1,97 @@
...
+
+
+# Function to replace single quotes and stringify text
+def replaces_quotes_and_stringify(text):
```
tahini, line 5: Do you need this function in other scripts? If so, you can put it in a separate file that you can include in the script here

tahini (changes requested):
Nice thanks!

There are some comments to make it more "production-ready" for Evolution. Some are comments for future improvement. But please add some TODOs to keep track in the code of those comments, if you plan to do them later.

generator/scripts/generateInputRange.py
samuel-duhaime: Thanks, that’s genius!

generator/scripts/generateInputRange.py
samuel-duhaime: Indeed, what a horror!

generator/scripts/generateInputRange.py
samuel-duhaime: Yes indeed, I use it in another file, so I will put it in another place.

generator/scripts/generateInputRange.py
samuel-duhaime: Yes, my goal is to do automated testing for Generator scripts, It's a good idea to catch the most common errors.

generator/scripts/generateInputRange.py
samuel-duhaime: Yes I will try to separate them better, thank you.

generator/scripts/generateInputRange.py
samuel-duhaime: Great idea.

generator/scripts/generateInputRange.py
samuel-duhaime: @tahini I'm not sure I understand what you have in mind.

generator/scripts/generateInputRange.py
davidmurray: Python has `os.linesep` for that:
https://docs.python.org/3/library/os.html#os.linesep

samuel-duhaime:
We are going to close, because it's now in pull request #364 .

PR #363, opened 2024-01-09
Add sendWelcomeEmail in createAndSave function

samuel-duhaime:
Add sendWelcomeEmail in createAndSave function

I put sendWelcomeEmail only in createAndSave, so it's only going to call if the user is new and the config is auth.welcomeEmail === true. 

I got server.json if the survey don't have a customServer file, but customServer file is the best way to implement the welcome email.

### DIFF
--- a/packages/evolution-backend/src/services/auth/participantAuthModel.ts
+++ b/packages/evolution-backend/src/services/auth/participantAuthModel.ts
@@ -8,6 +8,8 @@ import { BaseUser } from 'chaire-lib-common/lib/services/user/userType';
 import dbQueries from '../../models/participants.db.queries';
 import { ParticipantAttributes } from '../participants/participant';
 import { AuthModelBase, NewUserParams, UserModelBase } from 'chaire-lib-backend/lib/services/auth/authModel';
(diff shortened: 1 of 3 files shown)


packages/evolution-backend/src/services/auth/participantAuthModel.ts
```diff
@@ -44,6 +46,11 @@ class ParticipantAuthModel extends AuthModelBase<ParticipantModel> {
         });
         const user = this.newUser(userAttribs);
 
+        // Send a welcome email to the new user if the server is configured to do so
+        if (config.auth?.welcomeEmail) {
```
tahini, line 50: You should put the `===true` if that is the only value you want to put, or use __booleish. Here, the string "false" would return true

tahini (changes requested):

packages/evolution-backend/src/services/auth/participantAuthModel.ts
samuel-duhaime: C'est bon, j'ai fait la correction. On peut merge. :)

PR #367, opened 2024-01-19
preferences: Add a HOC component to make sure preferences are loaded

tahini:
The `WithPreferencesHoc` component loads the preferences upon first mount, then displays the wrapped component. A loading page is shown while preferences are loading.

The Survey and ValidateSurvey components use this HOC.


kaligrafy (approved):

kaligrafy:
lint fails

PR #369, opened 2024-01-29
audits: Use new audit table for audit filter

tahini:
The audit filter uses a subquery to the sv_audits table instead of the audits field of the interviews table.

The list of validation errors come from the sv_audits table as well.

Sequential tests are updated accordingly.


kaligrafy (approved):

tahini:
This PR adds a subquery to the validation list query. I expect performances will be degraded a bit in our production surveys, but by how much, I am not sure. @kaligrafy, if you have the opportunity, can you try this PR on the complete database and confirm that performances are not so bad, or au contraire, it's impossible to run?

tahini:
I'll merge and when/if performance becomes a problem, we can revisit the queries.

PR #370, opened 2024-02-05
add isInfo to audits

kaligrafy:


packages/evolution-backend/src/models/migrations/20240205090500_addInfoAuditTypeToAuditTbl.ts
```diff
@@ -0,0 +1,21 @@
...
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { Knex } from 'knex';
+
+const tableName = 'sv_audits';
+
+export async function up(knex: Knex): Promise<unknown> {
+    return knex.schema.alterTable(tableName, (table: Knex.TableBuilder) => {
+        table.boolean('is_info');
```
tahini, line 13: Can an audit be both info and warning? Usually, concepts like that will use a `level` field: error, warning, info (critical, debug, trace, but we don't need those for audits I suppose)

tahini (changes requested):
info/warning looks like a log's level concept. We should rather have a `level` field, with an enum or number to identify an audit level. The number allows to have a hierarchy of the levels. Otherwise, we can have an audit that can be both info and warning and that is not conceptually right.

packages/evolution-backend/src/models/migrations/20240205090500_addInfoAuditTypeToAuditTbl.ts
kaligrafy: Merging warning and error into a level would mean long refactoring. Let's keep as is for now, and add an issue for later?

kaligrafy:
Replaced by https://github.com/chairemobilite/evolution/pull/372

PR #372, opened 2024-02-05
add level column to audits

kaligrafy:


kaligrafy:
Will break old surveys using audits.

packages/evolution-backend/src/models/migrations/20240205101500_addLevelToAuditTbl.ts
```diff
@@ -0,0 +1,24 @@
...
+ * License text available at https://opensource.org/licenses/MIT
+ */
+import { Knex } from 'knex';
+
+const tableName = 'sv_audits';
+const columnName = 'level';
+
+export async function up(knex: Knex): Promise<unknown> {
+    return knex.schema.alterTable(tableName, (table: Knex.TableBuilder) => {
+        table.enu(columnName, ['error', 'warning', 'info']).defaultTo('error').notNullable().index();
```
tahini, line 14: Before dropping the is_warning column, you should set the level column to 'warning' if is_warning is true

packages/evolution-common/src/services/audits/__tests__/AuditUtils.test.ts
```diff
@@ -74,35 +74,35 @@ describe('auditsToAuditArray', () => {
...
+            level: 'error' as ('error' | 'warning' | 'info' | undefined),
             ignore: true,
             objectType: 'interview',
             objectUuid: arbitraryUuid
         }, {
             version: 1,
             errorCode: 'test-error-2',
             message: 'Test error message',
-            isWarning: false,
+            level: 'error' as ('error' | 'warning' | 'info' | undefined),
```
tahini, line 99: simply `as const` will compile properly, no need to cast to the whole enum

tahini (changes requested):
So you decided to do the "long" refactoring after all? ;-)

packages/evolution-common/src/services/audits/__tests__/AuditUtils.test.ts
kaligrafy: fixed

packages/evolution-backend/src/models/migrations/20240205101500_addLevelToAuditTbl.ts
kaligrafy: Not needed since only one survey uses audits v2 (od_mtl_2023) and audits needs to be reset anyway since they are not yet complete.

tahini (approved):

PR #373, opened 2024-02-06
include survey objects when auditing

kaligrafy:
frontend will need to recreate survey objects for stats and summary, so we need to get a serialized version from the server during audit

### DIFF
--- a/packages/evolution-backend/src/api/survey.validation.routes.ts
+++ b/packages/evolution-backend/src/api/survey.validation.routes.ts
@@ -38,7 +38,7 @@ router.get(
                         await copyResponsesToValidatedData(interview);
                     }
                     // Run audits on the validated_data
-                    const audits = await Audits.runAndSaveInterviewAudits(interview);
+                    const objectsAndAudits = await Audits.runAndSaveInterviewAudits(interview);
                     // TODO Here, the responses field should not make it to frontend. But make sure there are no side effect in the frontend, where the _responses is used or checked.
                     const { responses, validated_data, ...rest } = interview;
                     return res.status(200).json({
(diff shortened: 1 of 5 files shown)


packages/evolution-backend/src/api/survey.validation.routes.ts
```diff
@@ -38,15 +38,16 @@ router.get(
                         await copyResponsesToValidatedData(interview);
                     }
                     // Run audits on the validated_data
-                    const audits = await Audits.runAndSaveInterviewAudits(interview);
+                    const objectsAndAudits = await Audits.runAndSaveInterviewAudits(interview);
```
tahini, line 41: You can receive `{ audits, ...rest }` here...

packages/evolution-backend/src/api/survey.validation.routes.ts
```diff
@@ -38,15 +38,16 @@ router.get(
...
                     // TODO Here, the responses field should not make it to frontend. But make sure there are no side effect in the frontend, where the _responses is used or checked.
                     const { responses, validated_data, ...rest } = interview;
                     return res.status(200).json({
                         status: 'success',
                         interview: {
                             responses: validated_data,
                             _responses: responses,
-                            auditsV2: audits,
+                            auditsV2: objectsAndAudits.audits,
+                            surveyObjectsAndAudits: objectsAndAudits,
```
tahini, line 49: ... and use `{ ...rest }` here, otherwise, you send the audits twice

tahini (changes requested):
Or remove the auditsV2 from the response... as you wish

tahini (approved):

PR #374, opened 2024-02-06
Update base survey objects for serialized data

kaligrafy:


packages/evolution-common/src/utils/DateUtils.ts
```diff
@@ -0,0 +1,14 @@
+// can return invalid dates: use isNaN on date.getTime() to check if it's valid
```
tahini, line 1: Missing copyright header

packages/evolution-common/src/services/baseObjects/BasePerson.ts
```diff
@@ -274,57 +249,6 @@ class BasePerson extends Uuidable implements IBasePersonAttributes, IValidatable
...
-        if (
-            dirtyParams.baseTrips !== undefined &&
-            (!Array.isArray(dirtyParams.baseTrips) || !dirtyParams.baseTrips.every((trip) => trip instanceof BaseTrip))
-        ) {
-            errors.push(new Error('BasePerson validateParams: baseTrips should be an array of BaseTrip'));
-        }
-
-        // Validate baseVehicles (if provided)
-        if (
-            dirtyParams.baseVehicles !== undefined &&
```
tahini, line 321: Where are those validations now?

kaligrafy: they were never used, this was base object composition, which should not happen. We will composed only custom objects in survey itself.

packages/evolution-common/src/utils/DateUtils.ts
tahini: Please add unit tests for this file, it's a util after all!

tahini (changes requested):

tahini (approved):

tahini (approved):

tahini (approved):

PR #379, opened 2024-02-09
Migrate Generator to generator packages

samuel-duhaime:
##Description
After testing the Generator from Evolution with the Telework Survey, I found many mistakes, most of which were related to imports. I also encountered a few errors with the types. I decide to migrate to a packages, so it will be easier to maintain at the logic of Generator.

I decided to organize Python files better using Poetry (similar to Yarn, but for Python), so we can have a poetry.lock and pyproject.toml. This will enhance collaboration.

It's not perfect, but it was working after that. I'll work on fixing all the TODOs and FIXMEs when I have the time (probably not this month).


packages/generator/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "engines": {
+    "node": ">=18.0.0"
+  },
+  "name": "generator",
```
tahini, line 5: You can name it evolution-generator, to have a less generic name

tahini (changes requested):
iiuc, the files in example/demo_survey/generator should not be there?

tahini (approved):

PR #385, opened 2024-02-16
add sortBy icons and indexes

kaligrafy:


packages/evolution-frontend/src/components/pageParts/validations/InterviewList.tsx
```diff
@@ -122,9 +124,35 @@ const InterviewList = <CustomSurvey, CustomHousehold, CustomHome, CustomPerson>(
...
+                                </span>
+                                <a
+                                    className={`faIconRight${
+                                        sortBy.length === 1 &&
+                                        sortBy[0].id === column.accessor.toString() &&
+                                        sortBy[0].desc !== true
+                                            ? ' _active-background _blue'
+                                            : ''
+                                    }`}
+                                    onClick={() => handleSortingChange(column.accessor)}
```
tahini, line 138: `handleSortingChange` changes the current sorting order of the accessor, but here you have an arrow that should be 'asc' and another that should be 'desc'. If you click twice on the down arrow, there should be no change. You may have to send a second parameter in handleSortingChange with the direction or sorting and don't update if the order is the same

packages/evolution-frontend/src/components/pageParts/validations/InterviewListComponent.tsx
```diff
@@ -172,7 +174,14 @@ const InterviewListComponent: React.FunctionComponent<InterviewListComponentProp
                             title={props.t('admin:persons')}
                         />
                     </React.Fragment>
-                )
+                ),
+                label: props.t('admin:HouseholdSize'),
+                enableSortBy: true
+            },
+            {
+                accessor: 'responses.household.carNumber',
```
tahini, line 182: Add a TODO that says this should be project specific, like on line 133.

tahini (changes requested):

packages/evolution-frontend/src/components/pageParts/validations/InterviewList.tsx
kaligrafy: fixed

packages/evolution-frontend/src/components/pageParts/validations/InterviewListComponent.tsx
kaligrafy: removed car number completely, not needed unless someone asks for it.

PR #399, opened 2024-03-13
Remove path from InfoText of Generator

samuel-duhaime:
# Update Generator

## Description
Remove path from InputText of Generator. We don't need path for this input.

### DIFF
--- a/packages/evolution-generator/README.md
+++ b/packages/evolution-generator/README.md
@@ -145,7 +145,7 @@ Widgets are the building blocks of your survey. They define the structure and in
 | [inputRange](#range) | Input range name for InputRange (optional)              | string? |
 | comments             | Additional comments for the question (optional)         | string? |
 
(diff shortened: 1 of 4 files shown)


kaligrafy (changes requested):
Can you explain why you removed the path attribute?

samuel-duhaime:
> Can you explain why you removed the path attribute?

Yes. This is not InputText.tsx but Text.js from evolution-legacy. There is no path. It just a Label question with no answer. So the path is useless in this case.

image

tahini:
> Yes. This is not InputText.tsx from evolution-frontend but Text.js from evolution-legacy. There is no path. It just a Label question with no answer. So the path is useless in this case.
> 
Then the commit message is wrong and probably the type's name too.

What we call Input* are input widgets, part of the Question component. Here, you are updating what should be the Text component, not the InputText.

You may fix it here, to more clearly differentiate the 2 concepts

kaligrafy:
We should rename this InfoText if it is not an input

kaligrafy (approved):

PR #400, opened 2024-03-14
accept validateParams audits in filter

kaligrafy:


packages/evolution-backend/src/models/interviews.db.queries.ts
```diff
@@ -376,11 +376,12 @@ const getRawWhereClause = (
 ): string | [string, string | boolean | number] | undefined => {
     // Make sure the field is a legitimate field to avoid sql injection. Field
     // is either the name of a field, or a dot-separated path in a json object
-    // of the 'responses' field. We should not accept anything else.
+    // of the 'responses' field, or an audit name for validateParams,
+    // which includes "-" and ":". We should not accept anything else.
```
tahini, line 380: If different fields have different regexes, we should not have only one regex here. Instead, we should make the match with `-` and `:` in the case of audits and with `.` in the case of responses. Otherwise, there may be eventual unexpected results

tahini (changes requested):

packages/evolution-backend/src/models/interviews.db.queries.ts
kaligrafy: added a check for field = audits

greenscientist:
@kaligrafy , @tahini what is going on here?

tahini:
Je ne comprends pas le lien entre le code et le commit message.

kaligrafy:
replaced by #1942

PR #410, opened 2024-03-28
Simplify generatorConfigs.yaml

samuel-duhaime:
# Pull request

## Justification
Sometimes, we don't want to use every script from Generator for a specific survey.
Also, it's going to be way easier to configs for the new user working with Generator.

## Description
Add enabled_scripts for every script of Generator
Take the sections from Excel and not from generatorConfigs.yaml 
Remove any useless variables from generatorConfigs.yaml 
Use a survey_folder_path to keep the generator survey consistent

## After
image

## Before
I can't even take a screenshot of the old configs, because it's too long (51 lines vs 13 lines).
image


tahini (approved):
Cool!

PR #420, opened 2024-04-04
Update generator readme

samuel-duhaime:
# Pull request

## Description
Update generator readme


samuel-duhaime:
I close because I already have another pull request with the same pull request but without conflicts...

PR #445, opened 2024-04-16
generator: Add twoColumns and containsHtml common properties

tahini:
The widgets can have 2 additional properties: `twoColumns` and `containsHtml` that can be defined in the excel file.

The row is passed to the widget generator function, which can then request to generate any additional properties. A function for common question widget properties is added to generate those two properties.

Also, they are removed from the base input types and put in the input types themselves.


tahini:
I could not test because I'm missing custom widgets to make it work with the excel file, but the errors seem all related to the missing widgets, the code seems ok (given the said columns don't exist yet)

kaligrafy (approved):

PR #449, opened 2024-04-19
generator: Fix postal code regex for canadian postal code

tahini:
The validation message says we need to be in Canada, but the regex corresponds to Quebec. This fixes for any canadian postal code. Surveys can fine-tune their territory handling afterwards.


tahini:
@kaligrafy, we'll have to update the survey with this PR after

kaligrafy (approved):

PR #451, opened 2024-04-19
generator: Let conditional value properly handle booleans

### LINKED ISSUE #450: generator: conditional with boolean values not working
If the value in the excel file is a condtional liek 'VRAI', this is what is produced by the generator:

```
path: 'home._homeOutOfTerritory',
comparisonOperator: '===',
value: 'True'
```

The value should be `true` and not `True`

tahini:
fixes #450

If the value is a boolean, it is written as such in the generated code


tahini:
@kaligrafy une autre à ajouter à l'enquête

kaligrafy (approved):

PR #472, opened 2024-04-29
db: Add an index on responses->>_isCompleted

tahini:
This field is often used for validation filter lists and for large surveys, it can slow the queries.


tahini:
@kaligrafy teste-la d'abord sur od_mtl pour voir si ça améliore les performances de la requête.

kaligrafy (approved):

greenscientist:
@tahini so ?

PR #473, opened 2024-04-29
Create widgets schema init

kaligrafy:


tahini (changes requested):
You may need to rebase on `main`.

Also, can you explain what to do with those widgets? How are we supposed to interpret it?

About the `sections` property, why set it here? I would imagine widgets as components that you can move around in your survey. Sections are for each survey developer to decide. For example, `homeCarnumberElectric` can go in the home section or end section, or a section of its own if one is interested in a household's vehicles, but the widget would be the same.

kaligrafy:
This schema is for analysis and planning only, that's why I put the sections in wich nthey appeared in old surveys.

tahini (changes requested):
Please put this in a separate directory and add a README explaining what is the purpose of this file.

tahini (approved):

PR #495, opened 2024-05-16
playwright: More improvements to UI testing

tahini:

* Add functions to detect journeys, visited places, persons and active object IDs in a new `SurveyObjectDetector.ts` file
* Add test to wait for some text to be visible (for cases, like visited place template, where a button click does not change the current page URL, but simply its state)
* Add tests to validate the presence of specific options for radio and select widgets
* Make sure widgets are in the viewport so screenshot shows the last location.

### DIFF
--- a/packages/evolution-frontend/tests/ui-testing/SurveyObjectDetectors.ts
+++ b/packages/evolution-frontend/tests/ui-testing/SurveyObjectDetectors.ts
@@ -1,0 +1,154 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+// This file contains functions to detect survey objects from the survey update
+// data for playwright tests
+
+// Regexes for specific survey objects
+const uuidRegex = /[0-9a-f-]{36}/g
+const personObjectKeyRegex = /^responses\.household\.persons\.([0-9a-f-]{36})$/
+const activePersonKeyRegex = /^responses\._activePersonId$/
+const journeyObjectKeyRegex = /^responses\.household\.persons\.[0-9a-f-]{36}.journeys.[0-9a-f-]{36}$/
+const activeJourneyKeyRegex = /^responses\._activeJourneyId$/
+const visitedPlaceObjectKeyRegex = /^responses\.household\.persons\.[0-9a-f-]{36}.journeys.[0-9a-f-]{36}.visitedPlaces.[0-9a-f-]{36}$/
+const activeVisitedPlaceKeyRegex = /^responses\._activeVisitedPlaceId$/
+
+// Store the survey object IDs
+let personIds: string[] = [];
+let journeys: { [personId: string]: string[] } = {};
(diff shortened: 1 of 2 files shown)


packages/evolution-frontend/tests/ui-testing/SurveyObjectDetectors.ts
```diff
@@ -0,0 +1,154 @@
...
+
+const replaceWithActiveJourneyId = (str: string): string => {
+    const activeJourneyStr = '${activeJourneyId}';
+    return str.includes(activeJourneyStr) && typeof activeJourneyId === 'string' ? str.replace(activeJourneyStr, activeJourneyId) : str;
+}
+
+const replaceWithActiveVisitedPlaceId = (str: string): string => {
+    const activeVisitedPlaceStr = '${activeVisitedPlaceId}';
+    return str.includes(activeVisitedPlaceStr) && typeof activeVisitedPlaceId === 'string' ? str.replace(activeVisitedPlaceStr, activeVisitedPlaceId) : str;
+}
```
samuel-duhaime, line 53: Any reason for 3 similar functions and not one? We can keep it like that, it's not a big deal. 

```
const replaceWithActivePersonId = (str: string): string => {
    const activePersonStr = '${activePersonId}';
    return str.includes(activePersonStr) && typeof activePersonId === 'string' ? str.replace(activePersonStr, activePersonId) : str;
}

const replaceWithActiveJourneyId = (str: string): string => {
    const activeJourneyStr = '${activeJourneyId}';
    return str.includes(activeJourneyStr) && typeof activeJourneyId === 'string' ? str.replace(activeJourneyStr, activeJourneyId) : str;
}

const replaceWithActiveVisitedPlaceId = (str: string): string => {
    const activeVisitedPlaceStr = '${activeVisitedPlaceId}';
    return str.includes(activeVisitedPlaceStr) && typeof activeVisitedPlaceId === 'string' ? str.replace(activeVisitedPlaceStr, activeVisitedPlaceId) : str;
}
```

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -89,26 +78,16 @@ test.afterAll(async ({ browser }) => {
...
-    const matchGroups = str.match(personIdRegex);
-    if (matchGroups !== null) {
-        const personIndex = Number((matchGroups[0].match(/\d+/g)as any)[0]);
-        newString = str.replace(personIdRegex, personIds[personIndex]);
-    }
-    return newString;
-};
-
-// Click outside to remove focus, just fake a click on the app's outer div
+// Click outside to remove focus, fake a click on the left of the screen, to avoid the page scrolling out of current viewport
```
samuel-duhaime, line 81: Oh, that was what was going on!

samuel-duhaime (approved):

packages/evolution-frontend/tests/ui-testing/SurveyObjectDetectors.ts
tahini: I'm currently adding the trips and segments. I think there are definite simplifications to do. That will be for the next next PR

PR #514, opened 2024-05-24
Base objects refactoring: Person, Place, Segment, Junction, WorkPlace, SchoolPlace, Routing

kaligrafy:


packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
```diff
@@ -70,25 +76,98 @@ export type RidesharingUser = YesNoDontKnowNonApplicable;
 
 // A person can have multiple occupations, like fullTimeStudent + partTimeWorker:
 export const occupationValues = [
```
samuel-duhaime, line 78: For occupationValues, 

I have a question. For one survey (MJ), we asked this question with multiple choices (InputCheckbox). The respondent can choose student and worker status. So do we need all these choices?

kaligrafy: These are all the possible values, it does not mean we ask them in an input. When we define each class, we must include every value that could be in each attributes. When we will create the input themselves, we can then choose/filter values accordingly.

packages/evolution-common/src/services/baseObjects/Segment.ts
```diff
@@ -0,0 +1,511 @@
...
+    'driverUuid',
+    'carType',
+    'vehicleOccupancy'
+];
+
+export const segmentAttributesWithComposedAttributes = [
+    ...segmentAttributes,
+    'origin',
+    'destination',
+    'transitDeclaredRouting',
```
tahini, line 40: I'm still unsure about having one field per mode declared here, there can be a variety, but if they are in composed attributes, that means they will be in a json data field anyway, so we can easily revisit this relationship later.

packages/evolution-common/src/services/baseObjects/Segment.ts
```diff
@@ -0,0 +1,511 @@
...
+    'arrivalTime',
+    'driver',
+    'driverUuid',
+    'carType',
+    'vehicleOccupancy'
+];
+
+export const segmentAttributesWithComposedAttributes = [
+    ...segmentAttributes,
+    'origin',
```
tahini, line 38: Shouldn't the origin and destination be part of the segment's main attribute and not an extended one?

packages/evolution-common/src/services/baseObjects/WorkPlace.ts
```diff
@@ -0,0 +1,118 @@
...
+import { Result, createErrors, createOk } from '../../types/Result.type';
+
+/*
+ * A work place is a place associated with a person
+ * where they go to work on a regular basis (usual work place)
+ */
+
+export const workPlaceAttributes = [
+    ...placeAttributes,
+    'parkingType',
```
tahini, line 22: How is this different from a school place? can't they just all be places with a type instead?

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
tahini: Why not allow an array then?

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
```diff
@@ -70,25 +76,98 @@ export type RidesharingUser = YesNoDontKnowNonApplicable;
...
+export type StudentType = (typeof fullTimePartTimeTypeNoValue)[number];
+
+export const schoolPlaceTypeValues = [
+    'onLocation',
+    'hybrid',
+    'remote'
+];
+export type SchoolPlaceType = (typeof schoolPlaceTypeValues)[number];
+
+export type WorkerType = (typeof fullTimePartTimeTypeNoValue)[number] | 'parentalOrSickLeave';
```
tahini, line 132: On peut être en maladie ou congé parental et être étudiant et non travailleur, non? Ici, ça semble assumer que parentalOrSickLeave implique travailleur

packages/evolution-common/src/utils/ConstructorUtils.ts
```diff
@@ -14,17 +14,25 @@ export class ConstructorUtils {
      * @template T - The type of the params object.
      * @param params - The parameters object containing the attribute values.
      * @param attributeNames - The array of attributes names.
+     * @param attributeWithComposedAttributes - The array of attributes that have composed attributes. composed attributes are dealt with separately
```
tahini, line 17: What is a composed attribute and why dealt with separately?
Je pense avoir mélangé "composed" et "custom" dans mon review plus haut...

tahini (commented):
Quelques questions ici et là.

Mais sinon, ça fait une bonne base, bien testée sur quoi bâtir!

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
kaligrafy: If we allow an array, it is a big change in survey data... This needs discussions. Let's keep it as a single value for now.

packages/evolution-common/src/utils/ConstructorUtils.ts
kaligrafy: composed = separate object, using another class, validated separately. custom = custom attribute, saved in a json field in the db.

packages/evolution-common/src/services/baseObjects/Segment.ts
kaligrafy: See below. Composed attributes are standalone objects from other classes and will be saved in their own table in the db. Custom attributes are just values specific to a custom survey and will be saved as a json field in db.

packages/evolution-common/src/services/baseObjects/WorkPlace.ts
kaligrafy: For now, this is quite similar, but later on, we will add more specific attributes, like number of student/number of employee, etc.

packages/evolution-common/src/services/baseObjects/Segment.ts
kaligrafy: it is a composed object (Place) Extended means standrad attributes + composed + custom
composed attributes =  attributes + composed

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
kaligrafy: Never seen occupation as an array in any survey. Using an array would mean long discussions... You can add an issue for the future though.

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
```diff
@@ -37,19 +37,25 @@ export const ageGroupValues = [
...
 export type AgeGroup = (typeof ageGroupValues)[number];
-// Most of the time at least age OR age group must be defined
 
 export const genderValues = ['female', 'male', 'custom'];
 export type Gender = (typeof genderValues)[number];
 
 export const yesNoDontKnowNonApplicableValues = ['yes', 'no', 'dontKnow', 'nonApplicable'];
 export type YesNoDontKnowNonApplicable = (typeof yesNoDontKnowNonApplicableValues)[number];
 
+export const yesNoDontKnowPreferNotToAnswerValues = ['yes', 'no', 'dontKnow', 'preferNotToAnswer'];
```
tahini, line 50: Ce genre d'attribut pourrait être dans un fichier attributeType générique, ça doit être utilisé ailleurs que dans PersonAttributes. À déplacer quand tu auras à le répéter dans un autre fichier.

packages/evolution-common/src/utils/ConstructorUtils.ts
```diff
@@ -14,17 +14,25 @@ export class ConstructorUtils {
...
+        if (attributeWithComposedAttributes === undefined) {
+            attributeWithComposedAttributes = attributeNames;
+        }
 
         for (const attribute in params) {
             if (attributeNames.includes(attribute)) {
                 // If the key exists in the attributes object, assign the value to the initialized attributes
                 initializedAttributes[attribute] = params[attribute];
-            } else {
+            } else if (!attributeWithComposedAttributes.includes(attribute)) {
```
tahini, line 35: Donc, les attributs dont le nom est dans attributeNames, donc excluant les composed attributes vont dans initializedAttributes. Et les attributs qui restent et ne sont pas composed vont dans initializedCustomAttributes. Où vont les composed attributes? Peut-être mentionner dans la doc de cette fonction ce qu'ils deviennent et/ou ce que le caller doit faire avec après.

tahini (changes requested):
Un peu de doc à ajouter pour les composed attributes. Sinon, c'est bon

packages/evolution-common/src/services/baseObjects/attributeTypes/PersonAttributes.ts
kaligrafy: moved

packages/evolution-common/src/utils/ConstructorUtils.ts
kaligrafy: Added doc

tahini (approved):

Issue #431, opened 2024-04-09
generator: Create the section directory if it does not exist


tahini:
Right now, if the section directory does not exist, the generator script throws an error.

samuel-duhaime:
We can close, pull request #517 fix this issue.

PR #517, opened 2024-05-30
Add generate_folders script for Generator

samuel-duhaime:
# Pull request

## Justification
I always need to create a section folder when I create a new one, and I get an error for that when I forget to do it. I think it should be done automatically. Let's get the survey production easy.

## Description
Generate sections and common folders for Generator.


tahini (approved):
Please add in the commit message that it fixes #431

PR #520, opened 2024-06-05
Routing use start/end instead of arrival/departure dates/times

kaligrafy:
to be consistent with all the other classes


kaligrafy:
use #521 instead

PR #545, opened 2024-06-15
generator: Validate that row column exists before assigning

tahini:
In python `row['something']` does not return None if 'something' does not exist in row, it throws an exception, so we need to make sure the fields are available.

Recently added columns caused problems with previous excel files, so the checks are added to those here.

TODO We should validate this for each column, not just the new ones.


samuel-duhaime (approved):

davidmurray:
@samuel-duhaime @tahini 
FYI, dict.get(...) serves exactly this purpose!

```python
>>> d = dict(test_key="test_value")
>>> d['missing_key']
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: 'missing_key'
>>> print(d.get('missing_key'))
None
>>> d.get('missing_key', 'backup_value')
'backup_value'
```

tahini:
@davidmurray cool! merci!

PR #551, opened 2024-06-17
widgets: Remove the modal related props

### LINKED ISSUE #540: Survey: Let the opened modal management be done by the corresponding widget
Currently, the opened modal window is managed by the `Survey` component, which requires the whole component to be redrawn to open a help/confirm/other modal window.

Each component that has a modal window should be responsible to handle the modal state, otherwise, it needs to check the modal window shortname to see if its modal is the one that is opened, adding complexity and extra props pretty much everywhere.

tahini:
fixes #540 
 
Includes 2 commits:

widget: Let each widget manage their confirm modal opening
    
The button component does not receive the opened window shortname, nor
the open/closeConfirmModal functions. It handles its own state.

Those props are not passed around anymore either, each widget having
confirm popups need to manage their own state.

** Breaking change **

In surveys, some templates open confirm popup when deleting elements.
These will have to be handled in the widget itself instead. See the
`VisitedPlacesSection.tsx` diff in the example/demo_survey from this
commit. The change implies

* To add a state with the confirm delete path.
* To replace the calls to `openConfirmModal`, by a `setState` with the path.
* To replace the calls to `closeConfirmModal`, by a `setState` with
  `null`.
* To replace the modal shortname comparison by a comparison between the
  state and the object path.

and

widget: Remove the open/closeQuestionModal

The `Question` widget apparently already handles the modal questions
display. Simply removing all the props, even from the individual input
does not appear to change anything in the survey behavior. These props
were not really useful, it seems.


samuel-duhaime (approved):
Attendre le review de @kaligrafy , mais ça semble bon.

kaligrafy (approved):

PR #552, opened 2024-06-18
inputRange: add a `includeNonApplicable` property

### LINKED ISSUE #539: Add "Not applicable" for the user for range input
# Issue

## Justification
In some of our surveys, we often ask radio input questions that are not applicable to the user. As a result, the user doesn't know how to respond in this situation.

## Description
The user should be able to click on 'not applicable'. We should add a parameter to the widget (notApplicable: boolean). 

## Screenshot
We could do something like this.

image

tahini:
fixes #539

To avoid the participant having to select a position for a question that does not apply to him, there should be a "Non applicable" option that the user can select that will disable the input range. The answer in this case, instead of a `number` will be a `string` with value `na`.

### DIFF
--- a/example/demo_survey/src/survey/widgets/end.tsx
+++ b/example/demo_survey/src/survey/widgets/end.tsx
@@ -343,6 +343,7 @@ export const householdSurveyAppreciation = {
     type: "question",
     path: "household.surveyAppreciation",
     inputType: "slider",
+    includeNotApplicable: true,
     datatype: "number",
     twoColumns: true,
     maxValue: 10,
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -194,6 +194,10 @@ export type InputRangeType = {
     formatLabel?: (value: number, lang: string) => string;
     labels?: I18nData[];
     trackClassName?: string;
+    /** Whether to include a 'not applicable' checkbox that will disable the input */
+    includeNotApplicable?: boolean;
+    /** An optional label for the 'not applicable' text. Only used if includeNotApplicable is true */
+    notApplicableLabel?: I18nData;
 };
 
 export type InputDatePickerType = {
--- a/packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
+++ b/packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
@@ -18,7 +18,7 @@ import { CommonInputProps } from './CommonInputProps';
 import { sortByParameters } from './InputChoiceSorting';
 
 export type InputCheckboxProps = CommonInputProps & {
-    value: string;
+    value: string[];
     /** Value of the custom field if 'other' is selected */
     customValue?: string;
     inputRef?: React.LegacyRef<HTMLInputElement>;
--- a/packages/evolution-frontend/src/components/inputs/InputRange.tsx
+++ b/packages/evolution-frontend/src/components/inputs/InputRange.tsx
@@ -15,15 +15,16 @@ import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import { InputRangeType } from 'evolution-common/lib/services/widgets';
 import * as surveyHelper from 'evolution-common/lib/utils/helpers';
 import { CommonInputProps } from './CommonInputProps';
+import InputCheckbox from './InputCheckbox';
 
 export type InputRangeProps = CommonInputProps & {
-    value?: number;
+    value?: number | 'na';
     inputRef?: React.LegacyRef<ReactInputRange>;
     widgetConfig: InputRangeType;
 };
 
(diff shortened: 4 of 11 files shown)


tahini:
See the screenshot in the issue for what it looks like. 

Do you think the text of the "Non applicable" is ok? should be overridable by the individual surveys? And if so, should each question be able to update it?

Is `includeNonApplicable` a good name for the property?

There may also be some aria to add somewhere to make sure the "Non applicable" is linked to the text of the question somehow.

kaligrafy:
The text should be overridable in the survey config

example/demo_survey/src/survey/widgets/end.tsx
```diff
@@ -343,6 +343,7 @@ export const householdSurveyAppreciation = {
     type: "question",
     path: "household.surveyAppreciation",
     inputType: "slider",
+    includeNonApplicable: true,
     datatype: "number",
```
samuel-duhaime, line 347: Mettre en anglais. includeNotApplicable 

Comme dit par @kaligrafy, peux-tu ajouter un notApplicableLabel?

packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -194,6 +194,7 @@ export type InputRangeType = {
     formatLabel?: (value: number, lang: string) => string;
     labels?: I18nData[];
     trackClassName?: string;
+    includeNonApplicable?: boolean;
```
samuel-duhaime, line 197: Mettre en anglais, includeNotApplicable.

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
```diff
@@ -18,7 +18,7 @@ import { CommonInputProps } from './CommonInputProps';
 import { sortByParameters } from './InputChoiceSorting';
 
 export type InputCheckboxProps = CommonInputProps & {
-    value: string;
+    value: string[];
```
samuel-duhaime, line 21: C'était une erreur d'un autre commit?

packages/evolution-frontend/src/components/inputs/InputRange.tsx
```diff
@@ -15,15 +15,16 @@ import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
...
 export type InputRangeProps = CommonInputProps & {
-    value?: number;
+    value?: number | string;
     inputRef?: React.LegacyRef<ReactInputRange>;
     widgetConfig: InputRangeType;
 };
 
 export const InputRange = (props: InputRangeProps & WithTranslation) => {
-    const [value, setValue] = React.useState<number | Range | undefined>(props.value);
+    const [value, setValue] = React.useState<number | Range | undefined | string>(props.value);
```
samuel-duhaime, line 27: On veut un string ou mettre 'na' ?

packages/evolution-frontend/src/components/inputs/InputRange.tsx
```diff
@@ -15,15 +15,16 @@ import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import { InputRangeType } from 'evolution-common/lib/services/widgets';
 import * as surveyHelper from 'evolution-common/lib/utils/helpers';
 import { CommonInputProps } from './CommonInputProps';
+import InputCheckbox from './InputCheckbox';
 
 export type InputRangeProps = CommonInputProps & {
-    value?: number;
+    value?: number | string;
```
samuel-duhaime, line 21: On veut un string ou mettre 'na' ?

packages/evolution-frontend/src/components/inputs/__tests__/InputRange.test.tsx
```diff
@@ -143,3 +143,71 @@ describe('Should correctly render InputRange with various parameters', () => {
     });
 
 });
+
+describe('Should correctly render InputRange with non applicable', () => {
```
samuel-duhaime, line 147: Mettre en anglais 'not applicable'

samuel-duhaime (changes requested):
Peux-tu rajouter dans le Generator, rajouter un row includeNotApplicable?

packages/evolution-frontend/src/components/inputs/InputRange.tsx
tahini: Bon point, je vais mettre 'na' seulement

tahini:
> Peux-tu rajouter dans le Generator, rajouter un row includeNotApplicable?

C'est fait

packages/evolution-frontend/src/components/inputs/InputCheckbox.tsx
tahini: C'est un autre commit entier dans cette PR qui corrige une erreur sur ce type qui existe depuis longtemps...

PR #572, opened 2024-06-27
generator: Use a dictionary to get section column data

tahini:
This avoids errors when the number of columns do not match the expected count.

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_section_configs.py
+++ b/packages/evolution-generator/src/scripts/generate_section_configs.py
@@ -58,14 +58,13 @@ def generate_section_configs(excel_file_path: str):
         # Iterate through each row in the sheet, starting from the second row
         for row_number, row in enumerate(rows[1:], start=2):
             # Get values from the row
-            (
-                section,
-                title_fr,
-                title_en,
-                in_nav,
-                template,
-                parent_section
-            ) = get_values_from_row(row, headers)
+            row_dict = dict(zip(headers, (cell.value for cell in row)))
+            section = row_dict.get("section")
+            title_fr = row_dict.get("title_fr")
+            title_en = row_dict.get("title_en")
+            in_nav = row_dict.get("in_nav")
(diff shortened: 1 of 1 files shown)


samuel-duhaime:
Hmm, I'm not sure if we should proceed with that. The get_values_from_row function was a good way to raise an error if "Error getting values from row" occurred.

Can you please remind me of the specific case that was causing issues? Perhaps we can work on fixing the get_values_from_row function.

packages/evolution-generator/src/scripts/generate_section_configs.py
```diff
@@ -58,14 +58,13 @@ def generate_section_configs(excel_file_path: str):
         # Iterate through each row in the sheet, starting from the second row
         for row_number, row in enumerate(rows[1:], start=2):
             # Get values from the row
-            (
-                section,
-                title_fr,
-                title_en,
```
samuel-duhaime, line 64: Hmm, I'm not sure if we should proceed with that. The get_values_from_row function was a good way to raise an error if "Error getting values from row" occurred.

Can you please remind me of the specific case that was causing issues? Perhaps we can work on fixing the get_values_from_row function.

samuel-duhaime (changes requested):

packages/evolution-generator/src/scripts/generate_section_configs.py
tahini: If the sheet had more columns than the ones expected, it would fail (like the case where I removed the groups from the section config, we absolutely need to update the files to remove this columns, which is not ideal).

The `get_headers` function above made sure the expected headers are there, this one should raise an error if one columns is missing. But if there is more information in the file, it should not raise anything. This commit fixes that.

samuel-duhaime: Okay, sounds good.

samuel-duhaime (approved):

PR #576, opened 2024-06-28
Add generate UI tests script (template) with Generator

samuel-duhaime:
# Pull request

## Justification
I didn't feel like writing 1200 lines of code for the UI testing of the Telework survey, so I created a script to make it easier. We can refine this script later, but for now, it's a template that we can duplicate. It may not work perfectly, but it should assist in replicating the tests logic.

This is an example of the 1200 generated lines of code for the Telework survey. We just need to input the desired value for the test afterward.

```typescript
/********** Tests introduction section **********/

/* Test radio widget: introduction_quelleOrganisation */
// Extract value from quelleOrganisationChoices choice
testHelpers.inputRadioTest({ context, path: 'introduction.quelleOrganisation', value: '?' });

/* Test radio widget: introduction_quelCentreINRS */
// Test conditional siINRSConditional
testHelpers.inputVisibleTest({ context, path: 'introduction.quelCentreINRS', isVisible: undefined });
// Extract value from quelleOrganisationINRSChoices choice
testHelpers.inputRadioTest({ context, path: 'introduction.quelCentreINRS', value: '?' });

/* Test string widget: introduction_quelCentreINRSAutre */
// Test conditional siINRSAutreConditional
testHelpers.inputVisibleTest({ context, path: 'introduction.quelCentreINRSAutre', isVisible: undefined });
testHelpers.inputStringTest({ context, path: 'introduction.quelCentreINRSAutre', value: '?' });
```

## Description
Add generate UI tests script (template) with Generator Generate tests folder
Add .ts extension to is_ts_file
Raise error if missing enabled_scripts


packages/evolution-generator/src/scripts/generate_survey.py
```diff
@@ -33,7 +34,28 @@ def generate_survey(config_path):
         # Get the data from the YAML file
         survey_folder_path = surveyGenerator["survey_folder_path"]
         excel_file_path = surveyGenerator["excel_file_path"]
-        enabled_scripts = surveyGenerator["enabled_scripts"]
+        enabled_scripts = surveyGenerator.get("enabled_scripts", [])
+
+        required_scripts = [
```
tahini, line 39: Required by who? and/or for what?

packages/evolution-generator/src/scripts/generate_survey.py
```diff
@@ -33,7 +34,28 @@ def generate_survey(config_path):
...
+            "generate_sections",
+            "generate_widgets_configs",
+            "generate_widgets",
+            "generate_conditionals",
+            "generate_choices",
+            "generate_input_range",
+            "generate_libelles",
+            "generate_UI_tests",
+        ]
+        missing_scripts = [
```
tahini, line 51: Does it mean that you force the generate_UI_tests script? What will happen when you run it out of the box in MJ or LD?

packages/evolution-generator/src/scripts/generate_UI_tests.py
```diff
@@ -0,0 +1,154 @@
...
+            # Check if we've moved to a new section
+            if section != current_section:
+                current_section = section  # Update the current section tracker
+                ts_code += f"/********** Tests {current_section} section **********/\n\n"
+
+            # Adjust path for widgets in groups using mappings or '?' for unknown groups
+            if group:
+                group_path_mapping = {
+                    "householdMembers": "household.persons.${personId[0]}"
+                    # TODO: Add more group mapping
```
tahini, line 106: They should not be manually added. Replace the TODO with something to get the group mapping. There's usually a widget that has the same name as the group and we can use its path here (replacing the thing between `${}` by whatever precedes in singular with ID... for later

tahini (changes requested):
Very cool! So this will produce a template file that you can then copy paste to just add the values?

iiuc, all the fields will be in the template, even those from groups that we don't necessarily fill by default in the tests? so you can always start from a copy of the template and do whatever needs to be done?

Just some comments on the comments, but if it works as you wish, that's a great improvement!

packages/evolution-generator/src/scripts/generate_survey.py
samuel-duhaime: If there were no scripts, I had an error, but without much information if you don't know what was going on. I might propose a better way of doing that. I could probably just send a default value of false, and there won't be any error.

packages/evolution-generator/src/scripts/generate_survey.py
samuel-duhaime: I will propose a more elegant solution on Tuesday.

packages/evolution-generator/src/scripts/generate_UI_tests.py
samuel-duhaime: Oh yeah, that's smart. I will add that on Tuesday.

samuel-duhaime: This TODO comment is done.

samuel-duhaime:
@tahini Okay, I made some changes. You can review it again. Thank you.

tahini (approved):

PR #579, opened 2024-07-03
dep: Upgrade/remove dependencies in evolution-legacy/demo_survey

tahini:


example/demo_survey/package.json
```diff
@@ -27,21 +27,6 @@
         "test:survey": "LOCALE_DIR=$(pwd)/locales npx playwright test"
     },
     "dependencies": {
-        "@babel/cli": "^7.10.5",
```
samuel-duhaime, line 30: We don't use Babel anymore?

samuel-duhaime (approved):

example/demo_survey/package.json
tahini: Not in demo_survey

PR #580, opened 2024-07-03
dep: Remove dependencies to bundle-loader and babel-loader

tahini:
They are not useful in the webpacks.


greenscientist (approved):
Seems legit

samuel-duhaime (approved):

PR #583, opened 2024-07-04
server: Add a deferred server update callback mechanism

### LINKED ISSUE #582: Add mechanism to asynchronously send data back to the user through the session memory
We can use the session data to store some server-side calculated data to send to the user at the next request. For example, in the long distance survey, we need to calculate scenarios and alternatives and send the response back to the user on the next request.

See the req.session here for documentation on how to add session data (https://expressjs.com/en/resources/middleware/session.html). The session is automatically saved when the request ends. We may need to manually save it if the calcualtions are done asynchronously.

Then, at each call to the updateInterview, we can add the new information to the response, so the frontend has access to it (and remove from the session store).

tahini:
fixes #582

Add an extra parameter for server update callbacks, so that callbacks
that have time-consuming operations (like routing calculations) can
register the operation, which can then execute without blocking the
current client update call.

The callbacks can return quickly empty or other values, but register
an operation, identified by a `opName` and `opUniqueId`. The `opName`
ensures that only one operation for that name will be considered. The
`opUniqueId` allows to cancel any previous operation with the same name.
The `operation` to register is a function that receives a `isCancelled`
function and returns a promise that will run in the background. The
promise should return the values by path to set in the interview. The
interview will be saved with the new values and those can be sent back
to the client at the next update call.

The deferred update callback will save the updated values in the user's
session and they will be added to the returned server values in the next
client response.

Example usage of this mechanism

```
field: 'accessCode',
callback: (interview, value, path, registerUpdateOperation) => {
    // Use the update operation to set the city to Calgary, but 5 seconds later
    registerUpdateOperation({
        opName: 'city',
        opUniqueId: 1,
        operation: () => new Promise((resolve) => {
            setTimeout(() => {
                resolve({
                    'home.city': 'Calgary'
                });
            }, 5000);
        })
    });
}
```

### DIFF
--- a/example/demo_survey/src/server/serverFieldUpdate.ts
+++ b/example/demo_survey/src/server/serverFieldUpdate.ts
@@ -10,7 +10,10 @@ import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 export default [
     {
         field: 'accessCode',
-        callback: (interview, value, path) => {
+        callback: (interview, value, path, registerUpdateOperation) => {
+            if (typeof value !== 'string') {
+                return {};
+            }
             // Sample server update: Move the home to Caraquet, NB if access code starts with 111 and the city is not set
             if (value.startsWith('111')) {
                 if (_isBlank(getResponse(interview, 'home.city'))) {
@@ -28,7 +31,33 @@ export default [
             } else if (value === '1133-1133') {
                 // Special access code to test url redirection
                 return [{}, 'https://github.com/chairemobilite/evolution/']
+            } else if (value.startsWith('222')) {
+                // Use the update operation to set the city to Calgary, but 5 seconds later
+                registerUpdateOperation({
+                    opName: 'city',
+                    opUniqueId: 1,
+                    operation: () => new Promise((resolve) => {
+                        setTimeout(() => {
+                            resolve({
+                                'home.city': 'Calgary'
+                            });
+                        }, 5000);
+                    })
+                });
+                // Use the update operation to set the province to Alberta, but 10 seconds later
+                registerUpdateOperation({
+                    opName: 'province',
+                    opUniqueId: 1,
+                    operation: () => new Promise((resolve) => {
+                        setTimeout(() => {
+                            resolve({
+                                'home.region': 'Alberta'
+                            });
+                        }, 10000);
+                    })
+                });
             }
+            
             return {};
         }
     }
--- a/packages/evolution-backend/src/api/survey.user.routes.ts
+++ b/packages/evolution-backend/src/api/survey.user.routes.ts
@@ -9,6 +9,7 @@
  * user's own survey
  * */
 import express, { Request, Response, Router } from 'express';
+import _get from 'lodash/get';
 import moment from 'moment';
 import { isLoggedIn } from 'chaire-lib-backend/lib/services/auth/authorization';
 import { UserAttributes } from 'chaire-lib-backend/lib/services/users/user';
@@ -99,6 +100,11 @@ export default (authorizationMiddleware, loggingMiddleware: InterviewLoggingMidd
                 const ip = (req as any).clientIp;
                 const timestamp = moment().unix();
 
+                // Get the current session data to send to client, then set to undefined again
+                const session = req.session;
+                const sessionClientPaths = session.clientValues?.updatedPaths;
+                delete session.clientValues;
+
                 const content = req.body;
                 if (!content.valuesByPath || !content.interviewId) {
                     console.log('Missing valuesByPath or unspecified interview ID');
@@ -107,8 +113,11 @@ export default (authorizationMiddleware, loggingMiddleware: InterviewLoggingMidd
                 const valuesByPath = content.valuesByPath || {};
                 const unsetPaths = content.unsetPaths || [];
 
-                if (unsetPaths.length === 0 && Object.keys(valuesByPath).length === 0) {
-                    return res.status(200).json({ status: 'success', interviewId: req.params.interviewUuid });
+                if (unsetPaths.length === 0 && Object.keys(valuesByPath).length === 0 && !sessionClientPaths) {
+                    return res.status(200).json({
+                        status: 'success',
+                        interviewId: req.params.interviewUuid
+                    });
                 }
 
                 const interview = await Interviews.getInterviewByUuid(content.interviewId);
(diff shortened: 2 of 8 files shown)


tahini:
@greenscientist This will be useful for LD's scenario calculations and would have been useful for OD's transit calculation if we had it back then.

packages/evolution-backend/src/api/survey.user.routes.ts
```diff
@@ -99,6 +100,11 @@ export default (authorizationMiddleware, loggingMiddleware: InterviewLoggingMidd
                 const ip = (req as any).clientIp;
                 const timestamp = moment().unix();
 
+                // Get the current session data to send to client, then set to undefined again
+                const session = req.session;
+                const sessionClientValues = session.clientValues?.valuesByPath;
+                delete session.clientValues;
```
greenscientist, line 106: Wondering, is there possible race condition here if we process 2 requests at the same time ?

greenscientist (commented):
I'd really wish we had an high level design/user document to explain those concepts. 

Also, security wise, is there any way this could be abused ? Any impact on performance in general ?

example/demo_survey/src/server/serverFieldUpdate.ts
```diff
@@ -10,7 +10,10 @@ import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 export default [
     {
         field: 'accessCode',
-        callback: (interview, value, path) => {
+        callback: (interview, value, path, asyncCallback) => {
+            if (typeof value !== 'string') {
+                return {};
```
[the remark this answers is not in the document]
tahini: It's an example from the demo_survey. Honestly, everything is quite sketchy there, but if you reset to field to empty, I think the UI will put it to null, so no, not an error path.

packages/evolution-backend/src/api/survey.user.routes.ts
tahini: Sure, there are many possible race conditions:

* 2 requests could trigger the long callback, which will both run together at the same time. Any of them could finish first and update the client value
* Here we set and delete the sessionClientValue at the beginning of the request. And it will only be saved at the end of it (after a few async call). In theory, the client should wait patiently for the request to return before doing anything else, but a quick participant could refresh the page, which would trigger some updates, which could cause another update of the values and maybe a second set of values will come back to the client, or this one. But the most common case would be that 2 requests would return the same data, which wouldn't be dramatic.
* Also, about the comment if the mechanism can be abused, yes indeed it can, since it's documented to be for long-lasting operation, a malefic participant could log in and trigger the conditions that cause this operation to be run and if the programmer of the callback does not plan for some cancelation/throttling mechanism, it could indeed cause denial of service on the server. But that is the case for any serverFieldUpdate callback with longer running operations, it's just that since the deferred callback returns to the client faster, the UI becomes responsive again faster and can be "abused" faster too (though one could achieve the same effect with many calls to curls or something in a script, so the responsive of the UI is not a condition), and is more sensible to bad programming/bad flow of survey (if the callback condition is met too frequently).

The most important is that whatever is saved in the survey responses matches what will be returned to the client. The race condition is not as much here (because the current happy path survey flow prevents 2 requests from the same participant at the same time) than during the data saving phase (the deferredSave in the interview.ts file, or in the deferredUpdateCallback below) if 2 or more long lasting operations fill the same data at the same time.

With the current, it's the survey programmer's responsibility to take care that the long running operations do not race because we don't know that a long lasting operation is running, we don't know that a deferred update callback will eventually be called.

We could instead have the server field update callback register the long lasting operation to a path, so that evolution could cancel the previous operation if another one for the same path occurs. That will reduce, though not totally eliminate the risk of race.

tahini: And now that I pressed send on the previous comment, I think another more race-mitigating approach would be that instead of saving in the session the full data to send to the client, we could save only the paths that were modified, and just before sending the responses back, we get the data for those paths directly from the recently loaded/saved interview. That would ensure we send the data that is currently in the interview.

In parallel, in the deferredUpdateCallback call below, we should reload the session and save it in a db transaction, to make sure we have the latest data and here too it should be saved right away, after deleting the clientValues to avoid overwriting the deferred callback data.

Or we implement mutexes of response paths!

tahini:
@greenscientist so given my lengthy reply, what action should be taken on this PR to go forward? Do you request changes of any sort, or do we live with the limitations and embrace the potential?

greenscientist:
> @greenscientist so given my lengthy reply, what action should be taken on this PR to go forward? Do you request changes of any sort, or do we live with the limitations and embrace the potential?

I feel like we should at least try to implement some of your idea, to hardened this a little bit. At a minimum, have something that detect that 2 instance of the same request have been requested and log accordingly. Maybe have a unique id for each request, so the client knows which one to check and handle it properly. Does the client want both instance or just the latest ? Things like that.

tahini:
So, after thought, what is going to be implemented is this:

Instead of a callback function, evolution will provide a "callfront" function ( I know, it's not a thing), by which the serverUpdateCallback mechanism will register a promise (or a function returning a promise) that is expected to return the updated values by path at the end. The callfront function will also receive a path identifying the current serverUpdateCallback and some kind of nonce object that, if equal to that a preceding call for the same path will not run again the promise, but if not equal, it will cancel the previous call (via a cancellation check function, with a mechanism similar to the routing calls in transition)and replace with the new one.

Also, in the session data, only the paths will be saved instead of the path/value pair, so that what will be returned to the client should be the last values available in the interview.

This should 1- reduce the risk of race conditions and 2- reduce the survey developers' burden by letting Evolution implement a maximum of logic to avoid lengthy operations taking up server resources unnecessarily.

tahini:
This PR was updated as advertised. See commit message and doc in `serverFieldUpdate.ts` from evolution-backend for details

example/demo_survey/src/server/serverFieldUpdate.ts
```diff
@@ -28,7 +31,33 @@ export default [
...
+                    })
+                });
+                // Use the update operation to set the province to Alberta, but 10 seconds later
+                registerUpdateOperation({
+                    opName: 'province',
+                    opUniqueId: 1,
+                    operation: () => new Promise((resolve) => {
+                        setTimeout(() => {
+                            resolve({
+                                'home.region': 'Alberta'
```
samuel-duhaime, line 54: Normally, the value starts with a lowercase letter (alberta, and not Alberta). Same for Calgary.

packages/evolution-backend/src/api/survey.user.routes.ts
```diff
@@ -119,15 +128,54 @@ export default (authorizationMiddleware, loggingMiddleware: InterviewLoggingMidd
...
+                                // Save the updated paths in the current user's
+                                // session data for asynchronous update to the
+                                // client, but make sure the interview id is the
+                                // same as the one is was saved for
+                                const currentSessionValues = reloadedSession.clientValues || {
+                                    interviewId: interview.uuid,
+                                    updatedPaths: []
+                                };
+                                const updatedPaths =
+                                    currentSessionValues.interviewId === interview.uuid
```
samuel-duhaime, line 149: Can you put more comments on this part?

packages/evolution-backend/src/api/survey.user.routes.ts
```diff
@@ -119,15 +128,54 @@ export default (authorizationMiddleware, loggingMiddleware: InterviewLoggingMidd
...
+                    // Fetch the values to send to the client from the interview
+                    const sessionClientValues =
+                        sessionClientPaths !== undefined
+                            ? sessionClientPaths.reduce((acc, path) => {
+                                acc[path] = _get(interview, path);
+                                return acc;
+                            }, {})
+                            : {};
                     if (retInterview.serverValidations === true) {
                         // TODO See if we can do a `res.redirect`. It does not work with a local server, as the browser gets CORS Missing Allow Origin messages
```
samuel-duhaime, line 170: Oh no, Not a CORS error. I hate those kind of errors... :(

samuel-duhaime (approved):

example/demo_survey/src/server/serverFieldUpdate.ts
tahini: In the demo_survey, these are text values entered by the user, so city and province start with uppercase. That would be right if this was a choice like in mj.

packages/evolution-backend/src/api/survey.user.routes.ts
tahini: I added some more meat to the comment of line 140, which explains the function's behavior

greenscientist (commented):
I'm curious, instead of adding an extra param to updateInterview, that need to be undefined most of the time, could we have a new function (updateInterviewWithCallback), which would take this param. It would make it more obvious that this is what we want to do. and move complexity to a separate fonction. (That would call the simple one)

greenscientist:
I hate this habit of JS of just passing a struct with a bunch of option for the parameters...

greenscientist (commented):
Ok, I think I might be getting to the thing that was bugging me: 
this is in fact not a callback at all. You cannot call back to the client, so it's really just an async update which will update some values later on. You name this value so the client knows when it appears. Is the "path" just a semaphore with a results/status information or it contains the actual value requested ?
Or I'm an completely wrong?

greenscientist (approved):
I will let it go for now. 

I think we *again* lack a clear definition of the API between the frontend and the backend.

PR #584, opened 2024-07-04
sections: Make sure the invalid widget has an ID before focusing

tahini:
Some widgets, like InputRange, do not have an ID (and cannot have one, as the library used does not allow to set it), but they can be invalid. To automatically scroll to the invalid question, we make sure the input has an ID before focusing on it. Otherwise, we use the other code path that just scrolls the page to make the widget visible.

### DIFF
--- a/packages/evolution-legacy/src/components/survey/Section.js
+++ b/packages/evolution-legacy/src/components/survey/Section.js
@@ -58,8 +58,8 @@ export class Section extends React.Component {
     if (!this.props.allWidgetsValid && this.props.submitted && this.props.loadingState === 0)
     {
       const invalidInputs = document.querySelectorAll('.question-invalid input');
-      if (invalidInputs.length > 0) {
-        // Focus on invalid input if found
+      if (invalidInputs.length > 0 && invalidInputs[0].id) {
+        // Focus on invalid input if found and it has an ID (inputRanges do not have IDs, the library used does not support it)
         const inputElement = document.getElementById(invalidInputs[0].id);
         if (inputElement) {
           inputElement.focus();


tahini:
@samuel-duhaime  This will fix https://github.com/chairemobilite/od_mj_2023/issues/313

greenscientist:
I'm curious, why we don't have an html ID for each question ? That would be something good to have no?

tahini:
> I'm curious, why we don't have an html ID for each question ? That would be something good to have no?

Evolution has an ID for each _question_. But the `input` widget associated with each widget sometimes come from an external library, like the react-input-range. And they do not support an `id` for their input element. It's out of our control

greenscientist:
> > I'm curious, why we don't have an html ID for each question ? That would be something good to have no?
> 
> Evolution has an ID for each _question_. But the `input` widget associated with each widget sometimes come from an external library, like the react-input-range. And they do not support an `id` for their input element. It's out of our control

I might be missunderstand, can we wrap each one with a div or something so we have something to anchor?

samuel-duhaime:
> I might be missunderstand, can we wrap each one with a div or something so we have something to anchor?

Yeah, can we try that before?

tahini:
We do have a div with an ID, but the if that is modified here puts the focus in the input so that you can edit it right away, like type the text in, which won't happen if you just anchor to the `div`, which is what the fallback (the else) does.

tahini:
(for more context, you can expand down the code in the review to see the `else` that is called when the ID is not there

samuel-duhaime (approved):
All right then. :)

PR #589, opened 2024-07-09
Frontend: Fix title header overflow

### LINKED ISSUE #448: Do not show username in header of participant app
A workaround solution in surveys is to add the `display: none` to the item-nav-user, but this element is still part of the page's html. It should be absent altogether from the page. But his needs a separate Header component (and maybe a complete separate handling of the page's layout? I'm not sure). Currently, the one from Transition is used.

### LINKED ISSUE #574: Frontend: Fix top menu on cell phone
# Issue

## Justification
We lose the trust of the respondent when they start the survey on their cell phones because the top menu is broken.

## Description:
We need to fix the top menu on cell phones. It seems unnecessary to display the user's name on the cellphone, and if the survey name is lengthy, we should consider using ellipsis (...) or a similar method to prevent the top menu from breaking. I'm not certain if this requires a fix with Transition or Evolution.

## Screenshot:
For iPhone 12
image

samuel-duhaime:
# Pull request
fix: #574 fix: #448

## Description
Fix title header overflow
Make header H1 show big on big screen and small on small screen so the header doesn't break
Remove the user name for survey but show on admin
fixes #574 fixes #448

### DIFF
--- a/packages/evolution-legacy/src/styles/survey/_header.scss
+++ b/packages/evolution-legacy/src/styles/survey/_header.scss
@@ -3,25 +3,38 @@ header {
   height: 1.8em;
   border-bottom: 1px solid $black-pale;
 
-
   .header__content {
     font-size: $font-size-small;
     align-items: flex-end;
     display: flex;
     justify-content: space-between;
     padding: $xxs-size;
+
+    // Ensures the title text in the top navigation bar doesn't overflow
     h1 {
       opacity: 0.6;
       margin: 0;
+      width: 100%;
+      max-width: 180px;
       font-size: $font-size-small;
       font-weight: 300;
+      white-space: nowrap; /* Prevents the text from wrapping to the next line */
+      overflow: hidden; /* Hides the overflow text */
+      text-overflow: ellipsis; /* Adds an ellipsis to indicate hidden text */
+
+      // Adjusts on larger screens
+      @media (min-width: 600px) {
+        max-width: 300px; 
+      }
     }
+
     ul {
       padding: 0;
       list-style: none;
       display: flex;
       margin: 0;
     }
+
     li {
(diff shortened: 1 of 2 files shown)


packages/evolution-legacy/src/styles/survey/_header.scss
```diff
@@ -10,18 +10,27 @@ header {
     display: flex;
     justify-content: space-between;
     padding: $xxs-size;
+
+    // Ensures the title text in the top navigation bar doesn't overflow
     h1 {
       opacity: 0.6;
       margin: 0;
+      width: 100%;
+      max-width: 180px;
```
tahini, line 19: On normal browsers is there a chance that the text will be shown with ellipsis even if there is enough space?

tahini (commented):
Just a question for normal browsers. Would you have screenshots with super long title on both a browser on big screen and mobile?

packages/evolution-legacy/src/styles/survey/_header.scss
samuel-duhaime: No, it will only show an ellipsis if there is not enough space.

samuel-duhaime: Yes, with this setup, after 180px. 

image

tahini (changes requested):
Would it be possible to limit the size to 180 px only in the case of smaller widths? or set the width of the right part, so that the left will automatically take whatever space there is? Or make the overflow not hidden if there's room?

This makes it much better that the current overflow in the survey, so we can accept it before the next release if we don't have time to make it better.

greenscientist:
@samuel-duhaime Something that we should finish??

tahini (approved):

PR #592, opened 2024-07-10
Add scripts to produce documentation using jsdoc

### LINKED ISSUE #591: doc: Use jsdoc to document Evolution
I experimented a bit with `jsdoc` to allow documenting the API of Evolution and its extension mechanisms. It uses the comments on the methods and classes to generate an HTML documentation. Since some parts of our code, especially the extensible ones, are documented and explaining using those tags, we could use `jsdoc` to produce a human readable documentation for survey developers to read.

This would also highlight the lacks of our doc and encourage us to better document our public facing classes and functions.

tahini:
fixes #591

Add yarn scripts to generate documentation for the `evolution-common` javascript code.

Make sure there are no empty returns on JsDoc, otherwise 'yarn jsdoc' won't work.
Use @typedef to identify a type that will be used directly inside JsDoc.

Ignore the api/docs folder inside Git.

TODO

* Document the typescript code instead, so that the types are also documented
* Document all the evolution packages, not just `common`
* Provide a higher level "glue" to document high level concepts that can then link to the jsdoc complete documentation of the packages.

To use this commit, you need to install `jsdoc` globally on your system with

`npm i -g jsdoc`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -133,3 +133,6 @@ example/demo_generator/locales
 example/demo_generator/references
 example/demo_generator/src
 example/demo_generator/tests
+
+## JSDoc API Documentations
+/docs/api/
--- a/package.json
+++ b/package.json
@@ -45,7 +45,8 @@
         "format:all": "yarn format && yarn format:python",
         "generate-migration": "knex migrate:make",
         "reset-submodules": "rimraf transition/ && git submodule init && git submodule update",
-        "generateSurvey:example": "yarn workspace evolution-generator run generateSurvey --config_path src/examples/generatorExampleConfigs.yaml"
+        "generateSurvey:example": "yarn workspace evolution-generator run generateSurvey --config_path src/examples/generatorExampleConfigs.yaml",
+        "jsdoc": "yarn workspace evolution-common run jsdoc"
     },
     "dependencies": {},
     "devDependencies": {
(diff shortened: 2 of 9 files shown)


tahini:
I spent a good 15 minutes on this, no more, there's a lot of improvements to do, but it's promising.

The idea comes from a comment by @greenscientist  on #583 to have a high-level idea of what's going on. And since I did try to document the code as much as possible, for extension mechanisms at least, it would be great that this doc can be made available somehow at a high level for others to see without reading the code.

samuel-duhaime:
Yes, this is really interesting, I could see myself looking at this when trying to find or understand a helper function for Evolution. This has so much potential to help us document our projects.

greenscientist:
Yes, that's nice. We'll need to find a way to publish it some where and link it into our static doc.

samuel-duhaime (approved):

samuel-duhaime:
Yes, this could be nice. Let's rebase and push this @tahini .

tahini:
See the TODO, it is more than a simple rebase and merge, but if we want it, I'll update this.

samuel-duhaime:
@tahini, Not sure how to make you a reviewer of your own PR.

tahini:
Ok, si vous considérez que c'est good enough pour commencer, tant mieux, on peut faire ça. Par contre, ça prendrait une section dans le readme pour expliquer comment générer la doc en ajoutant le package jsdoc. Je ne l'avais plus d'isntaller, une chance que c'était dans le commit message, mais il faudrait l'ajouter au readme principal pour le futur (avec un message sur les limitations actuelles de la génération, ie evolution-common seulement).

PR #602, opened 2024-07-18
survey: Add helper functions for odSurvey and additional responses field

tahini:
This adds various helper function taken from surveys, for self declared persons, trips getters, visited places name getter and segments getters. Add tests for all these functions.

Also add object fields that may be present for those helpers.

@samuel-duhaime  @kaligrafy This commit was previously part of my 15-min segment section branch, but I need them for longue distance too (to get those functions out of the helper file that imports frontend data and thus cannot run in the backend)

### DIFF
--- a/packages/evolution-common/src/services/odSurvey/helpers.ts
+++ b/packages/evolution-common/src/services/odSurvey/helpers.ts
@@ -5,8 +5,19 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 
+import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import { getResponse } from '../../utils/helpers';
-import { Household, Journey, Person, Trip, UserInterviewAttributes, VisitedPlace } from '../interviews/interview';
+import {
+    Household,
+    Journey,
+    Person,
+    Segment,
+    Trip,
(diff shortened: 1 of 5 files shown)


tahini:
Wait before reviewing, I'll add a visited place geography getter to these

tahini:
@kaligrafy  @samuel-duhaime  this is ready for review

kaligrafy (approved):

packages/evolution-common/src/services/odSurvey/helpers.ts
samuel-duhaime, line 1: This is the best opportunity for us to pass Objects as parameters for helper functions before we implement them on our survey. Especially since we are often tempted to add parameters to functions, which eventually causes the function to become unreadable.

It's really hard to read and review functions like this: 
```typescript
helperFunction(placeNewName, undefined, undefined, interview). 
```

This is much simpler to read and review.
```typescript
helperFunction({ visitedPlace: placeNewName, interview })
```

tahini: You're right, especially that no survey yet uses those functions officially... I'll open an issue with your comment and update the whole file in a separate PR.

And that sounds like a python programmer comment ;-)

PR #605, opened 2024-07-19
od helpers: Replace arguments with a single object

### LINKED ISSUE #603: OD helper: Use objects as argument instead of a series of unnamed ones
Before they become officially API and require breaking changes to update, now's a good time to replace series of argument by an object, that makes it clearer to call and easier to add/remove them.

** original comment starts here

This is the best opportunity for us to pass Objects as parameters for helper functions before we implement them on our survey. Especially since we are often tempted to add parameters to functions, which eventually causes the function to become unreadable.

It's really hard to read and review functions like this: 
```typescript
helperFunction(placeNewName, undefined, undefined, interview). 
```

This is much simpler to read and review.
```typescript
helperFunction({ visitedPlace: placeNewName, interview })
```

_Originally posted by @samuel-duhaime in https://github.com/chairemobilite/evolution/pull/602#discussion_r1683416496_

tahini:
fixes #603

All new OD survey helpers take an object as argument, instead of positioned ones. This will ensure uniformity, as some helpers to come will have more arguments. It also simplifies specifying optional arguments.

The helper functions that were aliases to the OD survey ones keep their previous signature, to reduce breaking changes as these can possibly be used by [un-tested/any-typed] surveys.

### DIFF
--- a/example/demo_survey/src/survey/sections.ts
+++ b/example/demo_survey/src/survey/sections.ts
@@ -338,10 +338,10 @@ export default {
       }
       else if (person.didTripsOnTripsDate === 'yes' || person.didTripsOnTripsDate === true)
       {
-        const journeys = odSurveyHelper.getJourneysArray(person);
+        const journeys = odSurveyHelper.getJourneysArray({ person });
         if (journeys.length >= 1) {
             const currentJourney = journeys[0];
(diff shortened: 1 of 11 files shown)


kaligrafy (approved):

samuel-duhaime (approved):
Yes, it is much better to understand sketchy cases now. Thank you.

PR #613, opened 2024-07-25
Routing: use transition public API for routing in surveys

tahini:
This adds a function that requests route calculations from the Transition public API

It also adds a routing-function agnostic function that surveys can use for calculation.

Transiton URL and username/password can be set in the survey configuration at application startup.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -33,6 +33,12 @@ PHOTON_OSM_SEARCH_API_URL=PHOTONOSMSEARCHAPIURL
 #SSL_PRIVATE_KEY=/path/to/privkey.pem
 #SSL_CERT=/path/to/sslcert.pem
 
+# Set the 3 following variables to be able to route with the Transition public
+# API, with data from a specific Transition instance
+#TRANSITION_API_URL=http://transition.example.org
+#TRANSITION_API_USERNAME=
+#TRANSITION_API_PASSWORD=
+
 ##############################################################
 # Mailing configuration, required for sending emails to users
 # strategy can be 'smtp' or 'sendmail'
--- a/packages/evolution-backend/src/config/projectConfig.ts
+++ b/packages/evolution-backend/src/config/projectConfig.ts
@@ -28,6 +28,28 @@ interface ProjectServerConfig {
      * individual audits. It is optional. If the survey does not require auditing, just leave blank.
      */
     auditInterview?: (attributes: InterviewAttributes) => Promise<SurveyObjectsWithAudits>;
+    /**
+     * Configuration of a Transition instance for route calculations. If not
+     * set, route calculations will not use the Transition public API for
+     * eventual route calculations.
+     */
+    transitionApi?: {
+        /**
+         * URL for the Transition API. Can be set by setting the
+         * TRANSITION_API_URL environment variable
+         */
+        url: string;
+        /**
+         * Username for the Transition API. Can be set by setting the
+         * TRANSITION_API_USERNAME environment variable
+         */
+        username: string;
+        /**
+         * Password for the Transition API. Can be set by setting the
+         * TRANSITION_API_PASSWORD environment variable
+         */
+        password: string;
+    };
 }
 
 export const defaultConfig: ProjectServerConfig = {
@@ -80,6 +102,22 @@ const projectConfig = Object.assign({}, defaultConfig);
  * @param config The project specific configuration elements
  */
 export const setProjectConfig = (config: Partial<ProjectServerConfig>) => {
+    // Try to set the transition API object from the environment variables.
+    // FIXME For now, it is either set in the config or in the environment
+    // variables. Consider a mix to support for example url in config and
+    // credentials in env.
+    if (!config.transitionApi) {
+        const transitionUrl = process.env.TRANSITION_API_URL;
+        const transitionUsername = process.env.TRANSITION_API_USERNAME;
+        const transitionPassword = process.env.TRANSITION_API_PASSWORD;
+        if (transitionUrl && transitionUsername && transitionPassword) {
+            config.transitionApi = {
+                url: transitionUrl,
+                username: transitionUsername,
+                password: transitionPassword
+            };
+        }
+    }
     Object.assign(projectConfig, config);
 };
 
--- a/packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
+++ b/packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+// This file contains route calculation utilities that can be used by the
+// survey. The should call chaire-lib's routing services to calculate the route.
+import { RoutingOrTransitMode } from 'chaire-lib-common/lib/config/routingModes';
+import { RouteCalculationParameter, RoutingTimeDistanceResultByMode } from './types';
+import projectConfig from '../../config/projectConfig';
+import { URL } from 'url';
+
+// A name for this calculation source
+const CALCULATION_SOURCE = 'transitionApi';
+
+// Cache the authentication token, as global variable for this file
+// FIXME Consider using a class to encapsulate the token and calculation method
+// TODO Handle token expiration. For now though, the evolution app will restart more often than the token expiration
+let token: string | undefined = undefined;
+
+const getTransitionUrl = (transitionUrl: string) =>
+    new URL(transitionUrl.startsWith('http') ? transitionUrl : `http://${transitionUrl}`);
+
+const getToken = async () => {
+    if (token !== undefined) {
+        return token;
+    }
+
+    const transitionUrlObj = getTransitionUrl(projectConfig.transitionApi!.url);
+    transitionUrlObj.pathname = '/token';
+    const response = await fetch(transitionUrlObj.toString(), {
+        method: 'POST',
+        headers: {
+            'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({
+            usernameOrEmail: projectConfig.transitionApi!.username,
+            password: projectConfig.transitionApi!.password
+        })
+    });
+
+    // FIXME Every request will try to get a token, even if the previous one
+    // failed. Should we cache the error and not retry? Or retry only after x
+    // time, possibly using an exponential backoff?
+    if (response.status !== 200) {
+        throw new Error(`Cannot get token from transition: ${response.status}`);
+    }
+
+    const tokenResponse = await response.text();
+    token = tokenResponse;
+    return token;
+};
+
+/**
+ * Calculate the routes for a list of modes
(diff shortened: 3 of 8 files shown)


tahini:
It's tested with a local Transition server with the OD longue distance survey. And it works.

tahini:
Le plus important à reviewer serait le fichier `types.ts` car c'est l'API que les enquêtes vont utiliser. Pour le reste, il y aprobablement plusieurs corner cases pas couverts, pour la plupart avec des FIXME ou TODO, mais ça fonctionne en gros, donc ce serait OK pour longue distance

packages/evolution-backend/src/config/projectConfig.ts
```diff
@@ -28,6 +28,25 @@ interface ProjectServerConfig {
...
+         */
+        url: string;
+        /**
+         * Username for the Transition API
+         */
+        username: string;
+        /**
+         * Password for the Transition API
+         */
+        password: string;
```
tahini, line 48: @greenscientist Comment pourrait-on configurer ce mot de passe pour l'instance? Présentement, ce serait dans le code source, dans le `server.ts` au démarrage du serveur qu'on setterait ça (donc disponible dans github pour la postérité). Serait-ce une variable d'environnement? ou qqc à feeder  somehow au démarrage de l'instance?

greenscientist: Variable d'environnement.

greenscientist: L'URL me dérange moins, mais le username ouais, ça serait bien. 
Tk, il faut surtout avoir moyen de setter tout ça de facon dynamique en dehors du build docker

tahini: Donc, ça peut soit être dans des variables d'environnement ou direct dans la config.

.env.example
```diff
@@ -33,6 +33,12 @@ PHOTON_OSM_SEARCH_API_URL=PHOTONOSMSEARCHAPIURL
 #SSL_PRIVATE_KEY=/path/to/privkey.pem
 #SSL_CERT=/path/to/sslcert.pem
 
+# Set the 3 following variables to be able to route with the Transition public
+# API, with data from a specific Transition instance
+#TRANSITION_API_URL=http://ld2024.transition.city
```
greenscientist, line 38: We should put something like example.transition.city in the comment, not an actual site.

packages/evolution-backend/src/config/projectConfig.ts
```diff
@@ -80,6 +102,19 @@ const projectConfig = Object.assign({}, defaultConfig);
  * @param config The project specific configuration elements
  */
 export const setProjectConfig = (config: Partial<ProjectServerConfig>) => {
+    // Try to set the transition API object from the environment variables
+    if (!config.transitionApi) {
+        const transitionUrl = process.env.TRANSITION_API_URL;
```
greenscientist, line 107: So we might want to set the URL in the config and the credential in ENV. I don't think this is possible that way ?
But we can keep it that way and say it's an all or nothing way to configure

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
```diff
@@ -0,0 +1,125 @@
...
+
+// This file contains route calculation utilities that can be used by the
+// survey. The should call chaire-lib's routing services to calculate the route.
+import { RoutingOrTransitMode } from 'chaire-lib-common/lib/config/routingModes';
+import { RouteCalculationParameter, RoutingResultByMode } from './types';
+import projectConfig from '../../config/projectConfig';
+import { URL } from 'url';
+
+// TODO Handle token expiration. For now though, the evolution app will restart more often than the token expiration
+let token: string | undefined = undefined;
```
greenscientist, line 16: is there a way to name or mark this variable to say that it's global ? Can be confusing when you read the code. 
Might also need a comment to say that.

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
```diff
@@ -0,0 +1,125 @@
...
+        headers: {
+            'Content-Type': 'application/json'
+        },
+        body: JSON.stringify({
+            usernameOrEmail: projectConfig.transitionApi!.username,
+            password: projectConfig.transitionApi!.password
+        })
+    });
+
+    // FIXME Every request will try to get a token, even if the previous one failed. Should we cache the error and not retry? Or retry only after x time?
```
greenscientist, line 39: We could have some form of exponential backoff mecanism for the retry. But at the amount of request that we expect for now, that's not really a problem. But we should make sure the error is logged (I don't know if the throw there will be loggued.

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
```diff
@@ -0,0 +1,125 @@
...
+                departureTimeSecondsSinceMidnight: parameters.departureSecondsSinceMidnight,
+                scenarioId: parameters.transitScenario
+            })
+        });
+
+        if (response.status !== 200) {
+            throw new Error(`Unsuccessful response code from transition: ${response.status}`);
+        }
+        const routingResponse = await response.json();
+        const travelTimeByMode: RoutingResultByMode = {};
```
greenscientist, line 88: I don't like the name of the travelTimeByMode variable. It's not just a travelTime that we return but a full route.

packages/evolution-backend/src/services/routing/types.ts
```diff
@@ -0,0 +1,53 @@
+/*
+ * Copyright 2024, Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
```
greenscientist, line 6: We did not had routing type before when we where calling trROuting? Or that code was all in the MTL survey side ?

packages/evolution-backend/src/services/routing/index.ts
```diff
@@ -0,0 +1,58 @@
...
+import projectConfig from '../../config/projectConfig';
+import routeWithTransitionApi from './RouteCalculationFromTransition';
+
+/**
+ * Calculate the routes for a list of modes
+ * @param modes The list of modes to calculate the routes for
+ * @param parameters The parameters for the route calculation
+ * @param withGeojson Whether to include the GeoJSON in the response
+ */
+export const calculateRoutesByMode = async function (
```
greenscientist, line 23: I know we only support Transition for now, but can we have the Transition stuff into a private function that is called by this public one? So that we know where to add a new one easily.

greenscientist (changes requested):
Just some comments to address either by changing code or adding comments.

packages/evolution-backend/src/services/routing/index.ts
tahini: It is the case, at least it is the intent. This is the public function that chooses the method to call. Unfortunately, in typescript, we cannot have private file. Technically someone should never import directly RouteCalculationFromTransition, but we cannot force that. Unless we put all the function for all methods in this file... but that sucks

packages/evolution-backend/src/config/projectConfig.ts
tahini: No, all or nothing

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
tahini: It will be logged by the main function. If we log it here, it will be logged twice

packages/evolution-backend/src/services/routing/index.ts
greenscientist: Ok, I thought that lot of it was specific for Transition

tahini: No, parameters needs to be validated for all methods.

packages/evolution-backend/src/config/projectConfig.ts
tahini: Added a FIXME to support a mix of both

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
tahini: Added a comment. And a FIXME that this file could be rewritten as a class, with the token being an attribute.

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
tahini: Added the exponential backoff in the comment

packages/evolution-backend/src/services/routing/types.ts
tahini: Mtl called the trRouting's summary call (during the survey). For the validations, we also do direct calls to trRouting, so we get the whole response. Here, this is a new, simplified, API to get time and duration. Eventually, if/when we need it, we could add more information, if necessary. We could have another function, similar to this one, that returns the whole data. Of course, it would re-use 90% of what's here, it's just the API that would change.

greenscientist: The name of the types/functions implies that we return the full results. If we are more constrainted, we need names that reflect that.

tahini: RoutingTimeDistanceResultByMode

tahini:
I added a source field to the results, so that if we use a fallback method in case of error or no routing, we can identify where the results comes from.

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
```diff
@@ -0,0 +1,135 @@
...
+
+// This file contains route calculation utilities that can be used by the
+// survey. The should call chaire-lib's routing services to calculate the route.
+import { RoutingOrTransitMode } from 'chaire-lib-common/lib/config/routingModes';
+import { RouteCalculationParameter, RoutingTimeDistanceResultByMode } from './types';
+import projectConfig from '../../config/projectConfig';
+import { URL } from 'url';
+
+// A name for this calculation source
+const CALCULATION_SOURCE = 'transitionApi';
```
greenscientist, line 16: We should have have as a type, so that we can use it as an Enum elsewhere.

packages/evolution-backend/src/services/routing/index.ts
```diff
@@ -0,0 +1,58 @@
...
```
greenscientist, line 23: The name is wrong right?  We do not calculate route, we would calculate route statistics ?

greenscientist (changes requested):
Still one question.

packages/evolution-backend/src/services/routing/RouteCalculationFromTransition.ts
tahini: I'm not sure it should be an enum. Evolution may not know all possible sources (for example in od_longue distance, we implement a fallback when there is error or no routing).

greenscientist (approved):

PR #620, opened 2024-08-01
Test: Added functions to test input verification

GabrielBruno24:
Test that input fields will not accept invalid character types 
Test that invalid string values will highlight the box in red 
Test that entering an invalid date will autocorrect it 
Test that clicking the continue button when not all answers are filled will not advance to the next page 
Test that clicking the continue button when age is under 18 will open a popup

### DIFF
--- a/packages/evolution-frontend/tests/ui-testing/testHelpers.ts
+++ b/packages/evolution-frontend/tests/ui-testing/testHelpers.ts
@@ -49,6 +49,7 @@ type RegisterWithoutEmailTest = (params: CommonTestParameters) => void;
 type RegisterWithEmailTest = (params: { email: Email; nextPageUrl?: Url } & CommonTestParameters) => void;
 type HasUserTest = (params: CommonTestParameters) => void;
 type SimpleAction = (params: CommonTestParameters) => void;
+type ContinueWithInvalidEntriesTest = (params: { text: Text; currentPageUrl: Url, nextPageUrl: Url } & CommonTestParameters) => void;
 type InputRadioTest = (params: PathAndValueBoolOrStr & CommonTestParameters) => void;
 type InputSelectTest = (params: PathAndValue & CommonTestParameters) => void;
 type InputStringTest = (params: PathAndValue & CommonTestParameters) => void;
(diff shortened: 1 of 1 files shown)


packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -585,6 +626,26 @@ export const inputDatePickerTest = ({ context, path, value }: PathAndValue & Com
     });
 };
 
+/**
+ * Input a date that is invalid (such as being in the future) and check that it gets autocorrected to a different value.
```
tahini, line 630: It does not really check that it is autocorrected to a different value, it just makes sure the value is not the requested one. What is the actual expected value? Do we know? Or is it empty? You can update the doc here

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -628,3 +689,44 @@ export const waitTextVisible = ({ context, text, isVisible = true }: { text: Pat
...
+
+/**
+ * Click the next page button when not all options are filled and check that we remain on the same page.
+ *
+ * @param {Object} options - The options for the test.
+ * @param {string} options.text - The text of the button to click.
+ * @param {string} options.currentPageUrl - The URL of the current page.
+ * @param {string} options.nextPageUrl - The URL of the page that the button would normally take us too.
+ */
+export const continueOnEmptyPage: ContinueWithInvalidEntriesTest = ({ context, text, currentPageUrl, nextPageUrl }) => {
```
tahini, line 709: I don't think we need a different helper for empty or invalid pages. Both cases actually represent a pageWithInvalidWidgets, no matter the invalidity reason (empty mandatory or erroneous data)

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -628,3 +689,44 @@ export const waitTextVisible = ({ context, text, isVisible = true }: { text: Pat
         }
     });
 };
+
+const continueOnInvalidPage: ContinueWithInvalidEntriesTest = async ({ context, text, currentPageUrl, nextPageUrl }) => {
+    const button = context.page.getByRole('button', { name: text });
+    await button.scrollIntoViewIfNeeded();
+    await button.click();
+    await expect(context.page.waitForURL(nextPageUrl, {timeout: 1000})).rejects.toThrow("Timeout")
```
tahini, line 697: It's a clever wait to kind of add a timeout before going to a next page. I wonder if we would have any other mechanism to realize that we did not go to the next page, like waiting for any widget with the invalid class?

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -628,3 +689,44 @@ export const waitTextVisible = ({ context, text, isVisible = true }: { text: Pat
...
+
+/**
+ * Click the next page button when an age under 18 is clicked, verifies that the appropriate popup appears, and close it.
+ *
+ * @param {Object} options - The options for the test.
+ * @param {string} options.text - The text of the button to click.
+ * @param {string} options.currentPageUrl - The URL of the current page.
+ * @param {string} options.nextPageUrl - The URL of the page that the button would normally take us too.
+ */
+export const continueWithUnder18Age: ContinueWithInvalidEntriesTest = ({ context, text, currentPageUrl, nextPageUrl }) => {
```
tahini, line 723: This sounds very specific. It is actually a goto next page with a popup that appears. It could happen for any kind of reason other than the under 18.

tahini (changes requested):
C'est bon pour les fonctions elles-mêmes. Par contre, il faudrait en renommer quelques-unes pour mieux représenter ce qu'elle font réellement et non un cas spécifique de leur utilisation.

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
GabrielBruno24: It seems to depend on the situation and context. For example, today the latest valid date is 01/08/2024. If I type 02/08/2024 in an empty box, it gets corrected to 02/08/2020. However, if I first input 01/08/2024, and then try to change it to 02/08/2024, it changes back to 01/08/2024.

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
GabrielBruno24: We could check that a specific widget gets an invalid class on top of us remaining on the page. The function I wrote is a more general one where we try to advance without filling anything, but we could also do something where we test a specific widget.

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
GabrielBruno24: Not sure what you mean here. Is there already a test that does something similar? I don't seen any class named pageWithInvalidWidgets

PR #642, opened 2024-08-09
Add validation test for radio inputs

GabrielBruno24:
Adds testHelper functions to verify the validation of radio widgets, as well as the textbox that appears when the last option is selected.

Also refactors a few functions to avoid code duplication


packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -379,12 +401,8 @@ export const inputStringTest: InputStringTest = ({ context, path, value }) => {
  */
 export const inputInvalidStringTypeTest: InputStringTest = ({ context, path, value }) => {
```
samuel-duhaime, line 402: Même commentaire.

samuel-duhaime (changes requested):

samuel-duhaime (approved):
Okay, c'est bon.

PR #675, opened 2024-09-10
Add a message when no auth method is there on the login page

samuel-duhaime:
# Pull request

## Description
Add a message when no auth method is there on the login page 
Note: We add this message for surveys that can only be authenficate with an access token from an invitation, for example with Leo panel.

## Screenshot
Login page
image

### DIFF
--- a/locales/en/auth.json
+++ b/locales/en/auth.json
@@ -52,6 +52,7 @@
   "CheckMagicEmail": "Connection information has been sent to the email address you requested. If you do not see it, make sure it is not in the spam.",
   "UsePasswordLessAuthMethod": "Use authentication by email link",
   "UseLocalLoginAuthMethod": "Use authentication with password",
+  "UseNoLoginAuthMethod": "Currently, the only login method is through invitation. No other log in options are available at this time.",
   "WhyLoginInfo": "",
   "AnonymousLoginExplanation": "You will not be able to continue an incomplete survey if your session expires or if you change browsers.",
   "AuthAnonymousTitle": "Anonymously",
--- a/packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
+++ b/packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
@@ -72,6 +72,30 @@ const authPageHOC = <P extends AuthPageProps & WithTranslation>(WrappedComponent
             setCurrentAuthMethod(authMethod);
         };
 
+        // Check if there are no auth methods except directToken
+        const hasNoAuthMethodsExeptDirectToken = React.useMemo(() => {
+            // Check if there are no auth methods
+            const hasNoAuthMethods = authMethods?.length === 0;
+
+            // Check if there are no auth methods in the config
+            const hasNoConfigAuth = !Object.entries(config?.auth || {}).some(([key, value]) => {
+                // Handle key separately
+                if (key === 'directToken') {
+                    // 'directToken' is not considered a login method with a widget, so return false
+                    return false;
+                } else if (key === 'passwordless') {
+                    // 'passwordless' is considered a login method with a widget, so return true
+                    return true;
+                } else {
+                    // For other keys like 'anonymous', 'google', 'facebook', etc., check if the value is true
+                    return value === true;
+                }
+            });
+
+            // Make sure that there are no auth methods and no auth methods in the config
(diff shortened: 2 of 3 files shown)


locales/en/auth.json
```diff
@@ -52,6 +52,7 @@
   "CheckMagicEmail": "Connection information has been sent to the email address you requested. If you do not see it, make sure it is not in the spam.",
   "UsePasswordLessAuthMethod": "Use authentication by email link",
   "UseLocalLoginAuthMethod": "Use authentication with password",
+  "UseNoLoginAuthMethod": "Currently, the only login method is through the LEO panel. No other login options are available at this time.",
```
tahini, line 55: C'est trop custom comme message, il ne faudrait pas mentionner le panel Leo ici. Genre "through invitation"

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
```diff
@@ -101,6 +101,16 @@ const authPageHOC = <P extends AuthPageProps & WithTranslation>(WrappedComponent
                                 isAuthenticated={isAuthenticated}
                             />
                             <div className="apptr__separator"></div>
+                            {/* This method is used when the respondent can only log in using an access token */}
+                            {currentAuthMethod === 'noLogin' && authMethods.includes('noLogin') && (
```
tahini, line 105: Il faudrait que cette méthode soit mutuellement exclusive à toutes les autres. Probablement automatique. Par exemple, si dans la config `auth` n'a ni passwordless, ni anonymous, ni facebook, ni google, ni... à part directToken, alors on utilise le 'noLogin'

tahini (changes requested):

locales/en/auth.json
samuel-duhaime: C'est fait.

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
samuel-duhaime: C'est fait.

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
```diff
@@ -72,6 +72,17 @@ const authPageHOC = <P extends AuthPageProps & WithTranslation>(WrappedComponent
             setCurrentAuthMethod(authMethod);
         };
 
+        // Check if there are no auth methods (except directToken)
```
tahini, line 75: You do not really check the directToken part. 

Also `passwordless` is not `true`, but an object.

You should rather filter the auth method for those that are `!== false` and do not `directToken` as key.

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
```diff
@@ -72,6 +72,17 @@ const authPageHOC = <P extends AuthPageProps & WithTranslation>(WrappedComponent
...
+        const hasNoAuthMethods = () => {
```
tahini, line 76: This can be a constant instead of function, the value will not change. You can use React.useMemo()

tahini (changes requested):
Have you tested with all kinds of config.js with/without configured auth methods? With `true`/`false` or just absent? Note that you need to restart the `yarn build:dev` anytime you change the config.js file, it won't be correctly picked up when updated (I think).

For the commit title, technically, you are not "adding an auth method", you are adding a warning when no auth method is available from the login page.

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
samuel-duhaime: Done.

packages/evolution-frontend/src/components/pages/auth/AuthPageHOC.tsx
samuel-duhaime: @tahini 
Done. 

First time using React.useMemo(). Do we do that for performance?

samuel-duhaime: Yes, I tested every possible method. There was one issue (https://github.com/chairemobilite/evolution/issues/678), it was probably there before this commit.

tahini (approved):

Issue #681, opened 2024-09-11
Run UI tests on github


GabrielBruno24:
Github does not currently run the playwright UI tests when making a pull request

PR #682, opened 2024-09-11
Add UI testing to workflow

GabrielBruno24:
The github workflow now runs the playwright ui tests.
Also fixes a broken test that wasn't caught by the pipeline because it does not run it. Fix: #681

### DIFF
--- a/.github/workflows/evolution.yml
+++ b/.github/workflows/evolution.yml
@@ -12,15 +12,35 @@ on:
 jobs:
   build-and-test:
     runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgis/postgis
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        env:
+          POSTGRES_USER: testuser
+          POSTGRES_PASSWORD: testpassword
+          POSTGRES_DB: testdb
     strategy:
       matrix:
         node-version: [18.x, 20.x]
     env:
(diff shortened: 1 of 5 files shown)


tahini:
What is the status of this PR? It would be nice ot have those tests running, since the current work does break a lot of stuff sometimes, having this extra check that the demo survey work would be great! What was missing? Can you try rebasing it? Or was there something else?

GabrielBruno24:
> What is the status of this PR? It would be nice ot have those tests running, since the current work does break a lot of stuff sometimes, having this extra check that the demo survey work would be great! What was missing? Can you try rebasing it? Or was there something else?

There was an issue with the google API key in this repo, with caused the map widget to not load properly when running the tests. @greenscientist and me tried fixing it for a little bit but we weren't able to, though I don't remember the exact reason. I can take a look at it once I'm done with my current PR but for now make sure to run the tests locally.

samuel-duhaime:
@GabrielBruno24 
Maybe, we can just remove the question with Google Maps in the demo (and open an issue for that). So we at least have a few UI tests, it's still better than none.

greenscientist:
Sinon, une autre option c'est de rouler les tests UI uniquement sur les commits et non sur les PR. C'est pas idéal, mais c'est mieux que rien. On attrape les problèmes à postériori.

greenscientist:
@GabrielBruno24 Could you move out the test fix into a separate PR, we could work on the github workflow separately

GabrielBruno24:
> @GabrielBruno24 Could you move out the test fix into a separate PR, we could work on the github workflow separately

@greenscientist I already did this a while ago. I guess I can just close this PR for now.

Issue #663, opened 2024-09-05
Improve postal code validation


GabrielBruno24:
Same issue as https://github.com/chairemobilite/od_nationale_2024/issues/231 for the base evolution repo. 
Instead of just checking that a postal code is 3 letter-number pairs, we should check that it does not have the letters that are considered invalid for a canadian postal code (see: https://en.wikipedia.org/wiki/Postal_codes_in_Canada#Number_of_possible_postal_codes).
We could also make variants of the postal code widget for specific regions.

PR #687, opened 2024-09-16
Improve postal code verification.

GabrielBruno24:
Change the regex for the postal code validation so that it doesn't accept letters not valid for canadian codes.
Also adds the case insensitive flag to shorten the regex and not repeat letters with different cases. 
Adds some more tests for the postal code to the input validation test. 
Fix: #663

### DIFF
--- a/packages/evolution-generator/src/common/validations.tsx
+++ b/packages/evolution-generator/src/common/validations.tsx
@@ -259,7 +259,10 @@ export const postalCodeValidation: Validations = (value) => {
             }
         },
         {
-            validation: !/^[A-Za-z][0-9][A-Za-z]( )?[0-9][A-Za-z][0-9]\s*$/.test(String(value)),
+            // To be valid in Canada, the postal code cannot have the letters D, F, I, O, Q, or U. It also cannot have W or Z in the first character.
+            // See: https://en.wikipedia.org/wiki/Postal_codes_in_Canada#Number_of_possible_postal_codes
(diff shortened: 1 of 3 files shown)


packages/evolution-generator/src/common/validations.tsx
```diff
@@ -259,7 +259,9 @@ export const postalCodeValidation: Validations = (value) => {
             }
         },
         {
-            validation: !/^[A-Za-z][0-9][A-Za-z]( )?[0-9][A-Za-z][0-9]\s*$/.test(String(value)),
+            // To be valid in Canada, the postal code cannot have the letters D, F, I, O, Q, or U. It also cannot have W or Z in the first character.
+            // See: https://en.wikipedia.org/wiki/Postal_codes_in_Canada#Number_of_possible_postal_codes
+            validation: !/^[abceghj-nprstvxy][0-9][abceghj-nprstv-z]( )?[0-9][abceghj-nprstv-z][0-9]\s*$/i.test(String(value)),
```
tahini, line 264: Technically, evolution is not a _canadian_ platform, so we should not have a country specific postal code validation here. But it was like that before. Maybe just add a FIXME here to say that instead of returning the validation, this can be a function, which takes the country as parameter and return the correct postal validation.

tahini (changes requested):
What does "Imporse" mean?

GabrielBruno24:
> What does "Imporse" mean?

Oops meant "improve"

kaligrafy (approved):

tahini (approved):

PR #695, opened 2024-09-23
adminExports: explode arrays of object in csv exports

### LINKED ISSUE #694: adminExport: arrays of object should be exploded in csv file
If an array in the responses contains objects (for example `scenarioData`: `[{id: 1, mode: 'car'}, {id: 2, mode: 'bus'}]`, the data in the csv is currently shown in one column named scenarioData, with `[Object]|[Object]`.

It should be exploded, as those arrays are usually of limited size and each interviews will have the same indices. In this case, there should be instead 4 columns: `scenarioData.0.id`, `scenarioData.0.mode`, `scenarioData.1.id`, `scenarioData.1.mode`.

tahini:
fixes #694

When a survey attribute is an array of complex types (object), instead of simply joining the values with `|`, we explode the paths to display each of its fields, ie `attribute.0.field1`, `attribute.0.field2`, `attribute.1.field1`, etc.

For arrays of simple type (strings, numbers, etc), the current behavior is kept.

### DIFF
--- a/packages/evolution-backend/src/services/adminExport/exportToCsvByObjectWorkerPool.ts
+++ b/packages/evolution-backend/src/services/adminExport/exportToCsvByObjectWorkerPool.ts
@@ -26,6 +26,13 @@ type AttributeAndObjectPaths = {
     arrayPaths: string[];
 };
 
+// Test whether this attribute is an array of objects
+const isArrayOfObjects = function (data: unknown): data is object[] {
+    return (
+        Array.isArray(data) && data.length > 0 && data.every((obj) => typeof obj === 'object' && !Array.isArray(obj))
+    );
+};
+
 const getNestedAttributes = function (
     parentAttribute: string,
     _object: { [key: string]: unknown },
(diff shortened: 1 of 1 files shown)


kaligrafy (approved):

packages/evolution-backend/src/services/adminExport/exportToCsvByObjectWorkerPool.ts
```diff
@@ -50,7 +50,26 @@ const getNestedAttributes = function (
             attributesAndObjectPaths.objectPaths.push(parentAttribute);
         }
         if (Array.isArray(_object[attribute])) {
-            attributesAndObjectPaths.arrayPaths.push((parentAttribute ? parentAttribute + '.' : '') + attributeRenamed);
+            // Is this an array of simple or complex types?
+            if (
+                _object[attribute].length > 0 &&
+                _object[attribute].every((obj) => typeof obj === 'object' && !Array.isArray(obj))
```
samuel-duhaime, line 56: Can you name the if in a variable like something isArrayComplexTypes ?

packages/evolution-backend/src/services/adminExport/exportToCsvByObjectWorkerPool.ts
```diff
@@ -75,6 +94,14 @@ const getPaths = function (parentPath, _object, attributes: string[] = []) {
         if (attribute === '_actions') {
             continue;
         }
+        if (
+            Array.isArray(_object[attribute]) &&
```
samuel-duhaime, line 98: Same as the other stuff. Can you store that in a variable?

 Why do you need to check if it's an array and not on the other one? Isn't it supposed to be the same as the if statement on lines 55-56?

tahini: Done, the if was extracted in a `isArrayOfObjects` function

samuel-duhaime (approved):
Okay. Good.

PR #710, opened 2024-10-02
errorHandling: Make client-side exception reporting more robust

tahini:
Add a else clause in the express route in case the exception is not sent

In the frontend, let the exception be of unknown type and handle it properly.

### DIFF
--- a/packages/evolution-frontend/src/services/errorManagement/errorHandling.ts
+++ b/packages/evolution-frontend/src/services/errorManagement/errorHandling.ts
@@ -35,7 +35,7 @@ const redirectToErrorPage = (history?: History) => {
  * available
  */
 export const handleClientError = (
-    error: Error,
+    error: unknown,
     { interviewId, history }: { interviewId?: number; history?: History }
 ) => {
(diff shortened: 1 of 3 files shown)


packages/evolution-frontend/src/services/errorManagement/errorHandling.ts
```diff
@@ -35,7 +35,7 @@ const redirectToErrorPage = (history?: History) => {
  * available
  */
 export const handleClientError = (
-    error: Error,
+    error: unknown,
```
samuel-duhaime, line 38: Il y a rien de mieux que unknown ? :(

greenscientist (approved):
sigh mais ok :)

samuel-duhaime (approved):

packages/evolution-frontend/src/services/errorManagement/errorHandling.ts
tahini: Non! C'est vraiment le best pour ce cas!

PR #726, opened 2024-10-11
remove ChangePasswordPage with associated auth route

kaligrafy:
this is unused and the route was commented. Not needed anymore


tahini (approved):
delete party!

PR #741, opened 2024-10-18
survey: scroll to invalid question instead of trying to focus

### LINKED ISSUE #740: survey: scrolling to invalid radio or checkboxes do not work on iOs
On safari on ipad and iphone, when an invalid question is a radio or checkbox, the browser does not scroll to the widget after clicking the button to go to next section.

tahini:
fixes #740

On Safari on iPhone and iPad, calling the `focus` method on input elements of type radio or checkbox do not focus on those elements, so if they are out of view, the participant may not see that there are invalid widgets.

Since those widget types cannot really "focus" anyway, we simply use the scrollTo position for all widgets except the text boxes, which work on all known platforms.

### DIFF
--- a/packages/evolution-legacy/src/components/survey/Section.js
+++ b/packages/evolution-legacy/src/components/survey/Section.js
@@ -58,8 +58,12 @@ export class Section extends React.Component {
     if (!this.props.allWidgetsValid && this.props.submitted && this.props.loadingState === 0)
     {
       const invalidInputs = document.querySelectorAll('.question-invalid input');
-      if (invalidInputs.length > 0 && invalidInputs[0].id) {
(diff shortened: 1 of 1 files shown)


tahini:
Do not merge too fast, needs more testing with more widget types and platforms, but it seems to work on safari/ipad and linux/firefox with radios, maps, checkboxes and textboxes.

kaligrafy (approved):

tahini:
Testé sur iPad et desktop et widgets radio, text, checkbox, select, range, map, button et multiselect. Tout fonctionne

samuel-duhaime:
Très bien, je me suis toujours demandé c'était quoi le problème avec IOs et nos enquêtes. Merci d'avoir découvert.

PR #749, opened 2024-10-22
Add `yarn build:admin:prod` to the workflow

### LINKED ISSUE #748: workflow: Add `yarn build:admin:prod` to the workflow
#746 fixes an issue that would never have occurred if we had had the `yarn build:admin:prod` in the workflow. It would have failed at that point.

tahini:
fixes #748

This step will detect any problem in the admin bundle compilation.


tahini:
Expected to fail now, should be rebased after #746 is merged

kaligrafy (approved):

PR #756, opened 2024-10-24
generator: Use types from evolution instead of defining them

tahini:
This makes sure the widget types are as expected and defined in evolution. It removes unused fields in some default inputs for some types.

Tested with generated survey, it may cause compilation errors, easy to fix, on custom widgets.

### DIFF
--- a/packages/evolution-common/src/services/widgets/WidgetConfig.ts
+++ b/packages/evolution-common/src/services/widgets/WidgetConfig.ts
@@ -10,12 +10,18 @@ import { IconProp } from '@fortawesome/fontawesome-svg-core';
 import { UserInterviewAttributes, InterviewResponsePath, InterviewResponses } from '../interviews/interview';
 import { ParsingFunction, I18nData, InterviewUpdateCallbacks, ParsingFunctionWithCallbacks } from '../../utils/helpers';
 import { CliUser } from 'chaire-lib-common/lib/services/user/userType';
-import { TFunction, i18n } from 'i18next';
+import { TFunction } from 'i18next';
 
 type IconData = {
     url: string | ParsingFunction<string>;
     size?: [number, number];
 };
+
+type WidgetSize = 'small' | 'medium' | 'large';
+// ChatGPT: This could represent the directional alignment options as it "is focused on the direction in which elements are aligned"
+type WidgetDirectionAlign = 'center' | 'left' | 'right';
+// ChatGPT: This would represent the spatial alignment options as it "pertains to the spatial orientation"
+type WidgetSpaceAlign = 'vertical' | 'horizontal' | 'auto';
 
 /**
  * Validation function, which validates the value with potentially multiple
@@ -47,7 +53,7 @@ export type InputStringType = {
     defaultValue?: string | ParsingFunction<string>;
     maxLength?: number;
     datatype?: 'string' | 'integer' | 'float';
-    size?: 'large' | 'small' | 'medium';
+    size?: WidgetSize;
     textTransform?: 'none' | 'capitalize' | 'uppercase' | 'lowercase';
     placeholder?: string;
     inputFilter?: (input: string) => string;
@@ -72,7 +78,7 @@ type BaseChoiceType = {
     iconPath?: string;
     conditional?: ParsingFunction<boolean | [boolean] | [boolean, unknown]>;
     color?: string;
-    size?: 'large' | 'small' | 'medium';
+    size?: WidgetSize;
 };
 export type ChoiceType = BaseChoiceType & {
     value: string;
@@ -106,7 +112,7 @@ export type InputCheckboxType = {
     datatype?: 'string' | 'integer' | 'float' | 'text';
     columns?: number;
     rows?: number;
-    alignment?: 'vertical' | 'horizontal' | 'auto';
+    alignment?: WidgetSpaceAlign;
     customAlignmentLengths?: number[];
 };
 
@@ -121,10 +127,11 @@ export type InputRadioType = {
     sameLine?: boolean;
     customLabel?: I18nData;
     customChoice?: string;
+    customPath?: string;
     datatype?: 'string' | 'integer' | 'float' | 'text' | 'boolean';
     columns?: number;
     rows?: number;
-    alignment?: 'vertical' | 'horizontal' | 'auto';
+    alignment?: WidgetSpaceAlign;
     customAlignmentLengths?: number[];
 };
 
@@ -167,7 +174,7 @@ export type InputButtonType = {
     inputType: 'button';
     choices: ChoiceType[] | ParsingFunction<ChoiceType[]>;
     hideWhenRefreshing?: boolean;
-    align?: 'center' | 'left' | 'right';
+    align?: WidgetDirectionAlign;
     isModal?: boolean;
     sameLine?: boolean;
 };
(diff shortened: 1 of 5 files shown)


packages/evolution-generator/src/common/validations.tsx
```diff
@@ -22,7 +22,7 @@ export const requiredValidation: Validations = (value) => {
 };
 
 // Optional question
-export const optionalValidation: Validations = () => [{ validation: false }];
+export const optionalValidation: Validations = () => [];
```
samuel-duhaime, line 25: Are you sure about that? Empty array for optionalValidation ? Will it works all the time?

packages/evolution-common/src/services/widgets/WidgetConfig.ts
```diff
@@ -238,7 +240,7 @@ export type InputMapFindPlaceType = InputMapType & {
     updateDefaultValueWhenResponded?: boolean;
 };
 
-export type QuestionWidgetConfig = {
+export type BaseQuestionType = {
     type: 'question';
     twoColumns?: boolean;
     joinWith?: string;
```
samuel-duhaime, line 246: What is joinWith by the way?

packages/evolution-generator/src/common/defaultInputBase.tsx
```diff
@@ -10,36 +10,53 @@ import config from 'chaire-lib-common/lib/config/shared/project.config';
...
+    InputStringType,
+    InputTextType,
+    TextWidgetConfig
+} from 'evolution-common/lib/services/widgets/WidgetConfig';
 
 // TODO: Make sure to add tests for these default inputs
 
 // Input Radio default params
-export const inputRadioBase: inputTypes.InputRadioBase = {
+export const inputRadioBase: Pick<BaseQuestionType & InputRadioType, 'type' | 'inputType' | 'datatype' | 'columns'> = {
```
samuel-duhaime, line 30: Nice to know, that we can use Pick.

packages/evolution-generator/src/common/defaultInputBase.tsx
```diff
@@ -10,36 +10,53 @@ import config from 'chaire-lib-common/lib/config/shared/project.config';
 import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import surveyHelper from 'evolution-legacy/lib/helpers/survey/survey';
 import * as surveyHelperNew from 'evolution-common/lib/utils/helpers';
-import * as inputTypes from '../types/inputTypes';
+import {
```
samuel-duhaime, line 13: So If I understand correctly, these are the same Types that are use directly in the Input Frontend code, so there won't have disparity between the Generator and the Frontend ?

packages/evolution-generator/src/common/defaultInputBase.tsx
samuel-duhaime, line 1: I'm not sure I understand everything. Why are some parameters removed, like initValue and multiple?

packages/evolution-common/src/services/widgets/WidgetConfig.ts
samuel-duhaime, line 1: If we migrate everything there, there is a lot of repeated text. For instance, the size parameters are often mentioned, but they are never defined in a type Size. I had something similar in inputTypes. It will be harder to maintain, but I agree that will need to migrate everything to the same place.

packages/evolution-generator/src/types/inputTypes.ts
samuel-duhaime, line 1: Do we still need inputTypes.ts if we migrate everything in WidgetConfig.ts ? Maybe it's unnecessary to have inputTypes.

samuel-duhaime (changes requested):
So does the goal of this PR, is to be able to maintain one file for the Input types, because right now we use 2 files ?

packages/evolution-generator/src/common/defaultInputBase.tsx
```diff
@@ -10,36 +10,53 @@ import config from 'chaire-lib-common/lib/config/shared/project.config';
...
 
 // Input Radio Number default params
-export const inputRadioNumberBase: inputTypes.InputRadioNumberBase = {
+export const inputRadioNumberBase: Pick<
+    BaseQuestionType & InputRadioNumberType,
+    'type' | 'inputType' | 'columns' | 'sameLine'
+> = {
     type: 'question',
     inputType: 'radioNumber',
-    datatype: 'integer',
```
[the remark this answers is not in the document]
tahini: It's implicity for radio numbers, we don't need it

packages/evolution-generator/src/common/validations.tsx
tahini: The `optionalValidation` could also be `undefined` but I'm not sure it would compile. With empty array, it works (see code in `Validation.ts`, it will return true). With the right type, the `{ validation: false }` does not compile as it expects an errorMessage

packages/evolution-common/src/services/widgets/WidgetConfig.ts
tahini: It's a way to visually merge 2 related (like the email and phone number widgets when the "agree to be contacted" is set. Like  this:

Capture d’écran du 2024-10-24 15-32-20

(but yeah, it's "poorly" documented...)

packages/evolution-generator/src/common/defaultInputBase.tsx
tahini: > So If I understand correctly, these are the same Types that are use directly in the Input Frontend code, so there won't have disparity between the Generator and the Frontend ?

yes indeed, the objective of this PR is to use only one set of types and make sure it matches the WidgetConfig's type, which are those actually used by the widgets. They are not all correctly or perfectly typed, but at least, we have only one set of types to update.

Anyway, the goal is to move the defaultInputBase data to evolution-common, but the next button action depends on an action that is still in evolution-legacy so we can't do that yet.

packages/evolution-generator/src/common/defaultInputBase.tsx
tahini: > I'm not sure I understand everything. Why are some parameters removed, like initValue and multiple?

Because they are not used anywhere in the widgets. I don't know why they were there, probably the widget that served as example still had them, but they were removed.

packages/evolution-generator/src/types/inputTypes.ts
tahini: > Do we still need inputTypes.ts if we migrate everything in WidgetConfig.ts ? Maybe it's unnecessary to have inputTypes.

I'd leave it for now, those types are shorter and rightly named, but we might want to move them as described in the `WidgetConfig.ts` file eventually.

packages/evolution-common/src/services/widgets/WidgetConfig.ts
tahini: > If we migrate everything there, there is a lot of repeated text. For instance, the size parameters are often mentioned, but they are never defined in a type Size. I had something similar in inputTypes. It will be harder to maintain, but I agree that will need to migrate everything to the same place.

Indeed some common parameters like that could be extracted. I'll see what is worth extracting from the WidgetConfig's types.

packages/evolution-generator/src/common/defaultInputBase.tsx
samuel-duhaime: Oh yeah, I had a few problems with this action button before, mostly compilation problems if I remember correctly.

packages/evolution-generator/src/common/defaultInputBase.tsx
samuel-duhaime: Oh wow! Okay, good to know.

tahini:
I extracted the common string types

samuel-duhaime (approved):
Okay, thank you.

PR #759, opened 2024-10-24
survey: The `startSetInterview` prop type does nto include the `history`

### LINKED ISSUE #758: Weird responses field added in the responses
Since last week's update, there are some new and unwanted fields in the responses. For example `responses.
_activeSection|responses.length|responses.action|responses.location`.

Bisecting to see what caused these fields to be added to the responses.

tahini:
fixes #758

The `startSetInterview` function received in the props receives only 3 parameters that do not include the `History` object. This one is automatically added by the `dispatch` call. This added the History data to the survey instead of the prefilled answers, so we have a lot of un-necessary fields in the interviews.

### DIFF
--- a/packages/evolution-frontend/src/components/pages/Survey.tsx
+++ b/packages/evolution-frontend/src/components/pages/Survey.tsx
@@ -50,7 +50,6 @@ export type SurveyProps = {
     startSetInterview: (
         activeSection: string | null,
         surveyUuid: string | undefined,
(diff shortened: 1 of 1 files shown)


tahini:
@kaligrafy ce bug a mis de la cochonnerie dans les entrevues de od_nationale. ça fonctionne, mais il y aura plein de champs inutiles...

kaligrafy (approved):

PR #762, opened 2024-10-29
sections: Extract widget component from Section and others

tahini:
To ensure that widget generation is always handled the same way and new features are not added in a single code path, a `Widget` component is added to display the high level widget (Text, Group, Button, InfoMap or Question). There is one version for end components (`Widget`) and one for widgets that are part of a grouped object (`InGroupWidget`), with additional pathPrefix and different widgetStatus location. 

Section templates in `demo_survey` are updated to use this new component.

Also type the SectionConfig and pass the sectionConfig to the Section widget instead of individual fields. This is a **breaking change** for surveys as section templates will also need to be updated to match the Section's props. See commit message for details.

### DIFF
--- a/example/demo_survey/src/survey/templates/SegmentsSection.tsx
+++ b/example/demo_survey/src/survey/templates/SegmentsSection.tsx
@@ -7,19 +7,16 @@
 import React                    from 'react';
 import _cloneDeep from 'lodash/cloneDeep';
 import _get                     from 'lodash/get';
-import { withTranslation }      from 'react-i18next';
+import { WithTranslation, withTranslation }      from 'react-i18next';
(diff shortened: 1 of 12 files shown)


tahini:
This is pre-requisite work to bring the segment section template to evolution and reduce to a minimum what is the "template" part of the "template" as currently, there's a lot of common code between the templates and the Section component that are not exactly the same just because they did not evolve at the same pace, with bug fixes in one not matched in the other.

kaligrafy (approved):

tahini:
@samuel-duhaime Cette PR contient des breaking changes. Attention à la prochaine mise à jour d'Evolution dans les enquêtes. Je vais attendre avant de merger si jamais tu veux reviewer.

samuel-duhaime (approved):

PR #768, opened 2024-11-04
Type the section `preload`'s call and other fixes

tahini:
* Type the section`preload`'s call and change the second and more arguments to a single named argument, to make it simpler for implementations to use use only the required values without worrying about positioning.
* Fix short mode labels
* Fix UI tests after renaming the `driverLicenseOwnership` widget.


kaligrafy (approved):

tahini:
@samuel-duhaime ici, il y a un autre breaking change pour les enquêtes en cours, si jamais tu mets à jour evolution...

PR #778, opened 2024-11-08
survey: Add an appContext to the SurveyContext type for generic local…

tahini:
…es context

The `appContext` value is passed as prop to the `runClientApp` function and set in the `SurveyContext`. The `HomePage` widget uses this as a context for the i18n string translations, so it is possible to fine-tune these named translation strings to specific "context". This value is `undefined` by default and can be anything, either coming from the url, or specific to an instance, etc. It is only used to personnalize translation strings to this "context".

The `HomePage` widget is turned to a function component instead of a class to allow the use the multiple react contexts.


tahini:
@kaligrafy Ceci va permettre de personnaliser le texte sur la page d'accueil poru l'enquête Sherbrooke

Issue #501, opened 2024-05-21
15-minutes survey: Provide a configurable trip segments section


tahini:

tahini:
Note for the reviews of the PRs involved: there should be no remaining labels using the fr/en syntax, all should be `t` functions.  All localizable strings should be in english, no more french in Evolution. Make sure all localization strings can be overridden in surveys with an array passed to the `t` function, with one string always defined and the other overridable with a `survey:<rest of the string>`

PR #779, opened 2024-11-11
segments: add a group widget for the segments

tahini:
This group is the main group for the `segments` object of the survey. It contains the button labels, as well as the segment widgets.

The demo_survey now uses this group widget instead of the main one.

TODO The group or its widgets is not yet configurable


tahini:
this is task 5 of #501

kaligrafy (approved):

PR #790, opened 2024-11-20
Switch builds to Node 22

greenscientist:
Use node 22 for CI and docker builds by default

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build Node app
-FROM node:18-bullseye
+FROM node:22-bullseye
 WORKDIR /app
 
 # TODO split package.json copy and yarn install to have some intermediary images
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build Node app
-FROM node:18-bullseye
+FROM node:22-bullseye
```
tahini, line 2: Why bookworm in transition and bullseye in evolution?

tahini (approved):

Dockerfile
greenscientist: Because of trRouting/OSRM dependency in Transition, it's harder to upgrade.

greenscientist: Oh, I guess we just did not update it

greenscientist: I'll add a commit to update to bookworm

tahini (approved):

samuel-duhaime (approved):

greenscientist:
I've also added the latest node in the lint step

PR #798, opened 2024-11-25
Add a github workflow that run the generator example

greenscientist:
It install poetry, install the necessary package and run the generator with the example survey


greenscientist:
closes: #794

tahini (approved):
Merci!

samuel-duhaime (approved):

PR #815, opened 2024-12-05
build(dep): Update transition to latest version and bump to 0.5.0

tahini:
Require node 22 from now on, as chaire-lib requires it.

Steps followed to do the upgrade:

* Update transition repo to latest `main`
* Bump all package versions used by the evolution packages to match transition's
* Copy transition's `yarn.lock` file to entirely replace evolution's
* Run `yarn` to get extra packages needed by evolution
* Make everything compile, lint and work

Most errors encountered after updates relate to stricter type validation. Those were fixed by either typing or casting `as any` where proper typing is not in place yet (like in demo_survey).

The `jest-fetch-mock` from evolution-frontend was replaced by a manual mock of the global fetch.

Test snapshots were updated, mostly for new icon classes and names.


kaligrafy (approved):

kaligrafy:
Nice!

tahini:
Mis à jour avec les derniers changements

kaligrafy (approved):

PR #817, opened 2024-12-11
Some housekeeping around the WidgetOperation file

tahini:
See individual commits for more details

helpers: let parseValue support undefined and geojson data types

package.json: add clean and compile script for evolution packages for easier development flow, without re-compiling/cleaning chaire-lib

widgetStatus: Remove the `modalIsOpened` from `WidgetStatus` as it is handled by the Question widget

widgets: remove the `index.ts` file from common's widget service as its content has moved elsewhere


kaligrafy (approved):
Nice!

Issue #823, opened 2024-12-17
split yarn format so that the python part is separated


greenscientist:
(Could also add a "yarn format all")

PR #824, opened 2024-12-17
Update format python

samuel-duhaime:
# Pull request
fix: #823 

## Description
Add yarn format:all and yarn format:python in the package.json


greenscientist (approved):
I guess it's ok

PR #827, opened 2024-12-17
Fix bug where parsing null value with type geojson would throw an error.

GabrielBruno24:
Because the type of null is considered an object in javascript, the function would then try to read one of its properties, throwing an error. Also adds a test to make sure the bug is not reproduced.


tahini (changes requested):
There's an unnecessary test to remove, otherwise it's good!

tahini (approved):

PR #847, opened 2025-02-04
Update Transition to latest, with React/redux upgrade

tahini:
Upgrade react and other related packages:
    
    * react: 16.14 => 19.0.0
    * react-dom: 16.14 => 19.0
    * react-markdown: 8.0.7 => 9.0.1
    * react-redux: 7.2.9 => 9.2.0
    * react-router-dom: 5.3.4 => react-router: 7.0.2
    * react-simple-captcha: 6.1.7 => 9.3.1
    * redux: 4.2.1 => 5.0.1
    * redux-thunk: 2.4.2 => 3.1.0
    * remark-gfm: 3.0.1 => 4.0.0
    * @testing-library/react: 10.4.9 => 16.1.0
    * history: not used anymore


tahini:
Leaving this here with a list of remaining TODOs for tomorrow, but participant app works, with all login methods

- [ ] Fix unit tests that still use deprecated approaches
- [ ] Update chaire-lib and transition where PrivateRoute forces a user, but it can be `null`. (Look at the ConsentedRoute here to rewrite the Public and Private routes in transition)
- [ ] Language routes (/en, /fr) do not work and seem to use the default one.
- [ ] Make sure evolution's datepicker works as the package was updated
- [ ] Test the admin app: search by access code, with access code in url and create new interviews
- [ ] Make sure logins and registration work in admin, as well as user lists and updates
- [ ] Make sure survey edition works in admin
- [ ] Make sure validations work in admin
- [ ] Make sure monitoring work in admin

tahini:
This is a big one.

I cannot guarantee that _everything_ works, but the checked elements above work, in admin and participant app, UI tests run, unit tests were converted to react/testing-library. It's been tested with the demo_survey app.

example/demo_survey/config.js
samuel-duhaime, line 1: Can you say in the commit, why we don't need detectLanguageFromUrl anymore?

example/demo_survey/src/survey/templates/VisitedPlacesSection.tsx
```diff
@@ -160,8 +164,8 @@ export const VisitedPlacesSection: React.FC<SectionProps & WithTranslation & Wit
             title     = {visitedPlaceDescription}
             onClick   = {props.interview.allWidgetsValid && person._uuid === _personId && !selectedVisitedPlaceId ? (e) => selectVisitedPlace(visitedPlace._uuid, e) : null}
           >
-            <div className="survey-visited-places-schedule-visited-place-icon">
-              <img src={`/dist/images/activities_icons/${visitedPlace.activity}_plain.svg`} style={{height: '2em'}} alt={visitedPlace.activity ? props.t(`survey:visitedPlace:activities:${visitedPlace.activity}`) : ''} />
+            <div className="survey-visited-places-schedulpropse-visited-place-icon">
```
samuel-duhaime, line 167: Is the name of the variable is schedulpropse ? There are no mistake?

packages/evolution-frontend/src/actions/Auth.ts
```diff
@@ -5,13 +5,14 @@
...
-import { History, Location } from 'history';
+import { Location, NavigateFunction } from 'react-router';
+import { Dispatch } from 'redux';
 import { login, redirectAfterLogin } from 'chaire-lib-frontend/lib/actions/Auth';
 import { _isBlank } from 'chaire-lib-common/lib/utils/LodashExtensions';
 import * as AuthBase from 'chaire-lib-frontend/lib/actions/Auth';
 
-export const startDirectTokenLogin = (history: History, location: Location, callback?: () => void) => {
-    return async (dispatch) => {
+export const startDirectTokenLogin = (location: Location, navigate: NavigateFunction, callback?: () => void) => {
```
samuel-duhaime, line 14: Can we make a breaking change for this function in the commit? The parameters are not in the same place.

samuel-duhaime (changes requested):

example/demo_survey/config.js
tahini: I will add it to the config properties, so it can be typed. And the default value will be `true` (for backward compatibility)

packages/evolution-frontend/src/actions/Auth.ts
tahini: Cette fonction est interne à evolution-frontend, les enquêtes ne devraient pas importer, je ne crois pas qu'il y ait de breaking change du point de vue des enquêtes.

packages/evolution-frontend/src/components/survey/InfoMap.tsx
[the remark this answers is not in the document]
tahini: > Petite question personnelle: Pourquoi tu import le JSX type dans tous les fichiers alors que ça marchait auparavant ?

La réponse très pragmatique: "parce que ça ne compile plus si on ne le fait pas". Maintenant, si ta question est plus philosophique à savoir ce qui a changé depuis la version 16 pour faire qu'on soit maintenant obligés de faire ça, je ne sais pas...

samuel-duhaime (approved):
C'est bon de mon bord.

kaligrafy (approved):

PR #848, opened 2025-02-05
auth: fix admin auth configuration

tahini:
commit 40bea0acf8ab6d4128d5696a56a82ce4a1545f39 removed the localLogin from the demo_survey application. But that configuration was used for the admin application.

A new configuration option, `adminAuth`, can be used to configure the admin login configuration. In webpack and in the server initialization, the `adminAuth` replaces the `auth` option of the configuration.


tahini:
Note that even before the mentioned commit, if the auth configuration did not include the localLogin, which is now unsupported, the admin login page would not have worked.

kaligrafy (approved):

PR #870, opened 2025-02-20
Add UI Test run on CI

greenscientist:
New github workflow that runs the playwright ui tests. Based on the main CI one, without the lint/unit-test/etc.
Start a postgresql DB to be able to run the application.

We only run on commit to have access to the Google Key secret

Extracted and adapted from PR #682

### DIFF
--- a/.github/workflows/ui-test.yml
+++ b/.github/workflows/ui-test.yml
@@ -1,0 +1,72 @@
+# Run the UI tests
+
+name: UI Tests
+
+on:
+  push:
+    branches: [ main ]
+  # We are not running on Pull Request, since we need the Google API Key
+
+jobs:
+  build-and-test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgis/postgis
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+        env:
+          POSTGRES_USER: testuser
+          POSTGRES_PASSWORD: testpassword
+          POSTGRES_DB: testdb
+    strategy:
+      matrix:
+        node-version: [22.x]
+    env:
+      PROJECT_CONFIG: ${{ github.workspace }}/example/demo_survey/config.js
+      PG_CONNECTION_STRING_PREFIX: postgres://testuser:testpassword@localhost:5432/
+      PG_DATABASE_PRODUCTION: testdb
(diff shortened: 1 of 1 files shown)


greenscientist:
Tested on my personnal repo and it ran to completion successfully. (With the API Key set)

samuel-duhaime (approved):

GabrielBruno24 (approved):

greenscientist:
I'm curious about why the step yarn test:ui:install-dependencies is not required here.

greenscientist:
It was the wrong version with the comment that were wrong.

.github/workflows/ui-test.yml
```diff
@@ -0,0 +1,72 @@
...
+      uses: actions/setup-node@v4
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: Install
+      run: yarn
+    - name: Compile
+      run: yarn compile
+    - name: Build Client bundle
+      run: yarn build:prod
+      # Skipped build:admin since we don't have UI for them at the moment0
```
samuel-duhaime, line 52: Small typo for moment0

samuel-duhaime (approved):

tahini:
> I'm curious about why the step yarn test:ui:install-dependencies is not required here.

I suppose the container that gets installed may already have those browsers installed? We can merge this and see if it correctly runs and revert or fix if there's any problem.

.github/workflows/ui-test.yml
```diff
@@ -0,0 +1,72 @@
+# Run the UI tests
+
+name: UI Tests
+
+on:
+  push:
+    branches: [ main ]
+  # We are not running on Pull Request, since we need the Google API Key
```
tahini, line 8: And why again do branches have access to that key, but not PRs? Someone could make a random PR just to run something against Google with our API key and close the PR?

tahini (approved):

.github/workflows/ui-test.yml
greenscientist: Action on Commit on the repo runs after an maintainer have pushed a commit and thus github consider this as safe and export the "secrets" to the github action environnement. 
THe same would be true if a PR was made from a branch in the local repo. 
But when we can a PR run from branch from random people on the internet (from various forks), github to not expose the secrets as a security measure. 
Does that answer the question?

greenscientist: In short, yes to your second question.

Issue #871, opened 2025-02-20
README: Update Playwright instruction


samuel-duhaime:
# Issue

# Description
Update UI Playwright instruction in README.

PR #872, opened 2025-02-20
doc: update Readme with correct instructions to run UI tests

tahini:
fix: https://github.com/chairemobilite/evolution/issues/871

The command has recently been changed to `yarn test:ui` to run the playwright tests, in coherence with most of the surveys and to make it clearer that it is UI tests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,13 +137,13 @@ yarn start
 
 Run the UI tests
 ```
-yarn test:survey
+yarn test:ui
 ```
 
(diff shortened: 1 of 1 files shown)


greenscientist:
Is the test:ui:install-dependencies useful to reference there?

tahini:
> Is the test:ui:install-dependencies useful to reference there?

What do you mean? There are instructions above that mentions that this should be run or the browser installs can be fine-tuned depending on the need.

greenscientist:
Ok, missed that on my first read.

greenscientist (approved):

PR #895, opened 2025-03-03
Restrict the build docker workflow to run only on the official repo

greenscientist:
We don't want it to run on people's forks

### DIFF
--- a/.github/workflows/build-docker-image.yml
+++ b/.github/workflows/build-docker-image.yml
@@ -15,6 +15,8 @@ env:
 jobs:
   # Branch push job - builds and pushes to both registries
   branch-push:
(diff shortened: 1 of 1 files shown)


samuel-duhaime (changes requested):
We might consider whether we should do the same for the UI tests workflow. I believe we are missing the Google API map password to run the tests on fork.

image

greenscientist:
@samuel-duhaime that's a different issue. For that it would be interesting if we could do something in the test itself to be able to skip the google map part and still run some of the test.

samuel-duhaime (approved):
Okay, I agree that the other problem should be another issue.

Issue #232, opened 2023-09-01
Enter key in login screen must trigger the email confirm button


kaligrafy:
Right now it just loops back to the page and nothing happened

millette:
Voir le PR https://github.com/chairemobilite/transition/pull/726 dans transition.

P.S.: je ne comprends pas trop pourquoi il y a un ```onKeyUp``` sur ```Button```.

PR #908, opened 2025-03-25
Update Transition submodule

samuel-duhaime:
# Pull request

## Description

This fixes the problems with the login and register page. 
When the user presses 'enter' at the good place, it won't go back to the consent page, and it will go to the survey. 
Fix: #232


greenscientist (approved):
Guessing that it's fine!

PR #907, opened 2025-03-19
Add a SectionProgressBar in the Section page.

samuel-duhaime:
# Pull request
Fix: #555 

## TODO

- [x] Add an UI test for SectionProgressBar
- [x] Add the UI test in the test template
- [x] Save the _completionRate inside the interview
- [x] Add a sectionProgressBar config inside the survey (hasSectionProgressBar ?), and only show the bar for the survey with this config.
- [x] Add CSS for the percentage like in Figma.

## Description
Add a SectionProgressBar in the Section page. 
Calculate and save the survey completion percentage when clicking on the 'Save and Continue' button.
Put the survey completion inside a circular-progress.
Only show SectionProgressBar if the config is hasSectionProgressBar for each section.
Add a sectionProgressBarTest UI test. Add these tests to the demo_survey.
Also generate the sectionProgressBarTest inside the UI template tests.
Format _survey.scss.
Format frontendHelper.test.ts and a responses._completionPercentage inside validateButtonActionWithCompleteSection

PR #910, opened 2025-04-01
Add a SectionProgressBar in the Section page. (#907)

samuel-duhaime:
# Pull request

- [x] Cherry-pick from 3e0fcf2
- [x] Cherry-pick bug with section progress and title

## Description
Calculate and save the survey completion percentage when clicking on the 'Save and Continue' button. Put the survey completion inside a circular-progress. Only show SectionProgressBar if the config is hasSectionProgressBar for each section. Add a sectionProgressBarTest UI test. Add these tests to the demo_survey. Also generate the sectionProgressBarTest inside the UI template tests. Format _survey.scss.
Format frontendHelper.test.ts and a responses._completionPercentage inside validateButtonActionWithCompleteSection
Cherry-pick of a314ff3

Run test on 0.3.0 branch 
Enable the basic workflow for the 0.3.0 branch
Cherry-pick from 3e0fcf2
Add the VisitedPlace type and _cloneDeep function in this commit, to fix compilation problems.

The section completion percentage will now start at 0% when the first section begins and will reach 100% once the last section is initiated.
To achieve this, we will disregard completed sections when calculating the total number of sections.
Test the completed page and the section progress bar in the demo survey.
Cherry-pick from dcf8790
Add a few types to solve the compilation problems.

Move the circular progress at the end. 
Move the circular progress at the end of the div so the user will read the title more easily and only focus on the completion rate if needed.
Cherry-pick from 02b84ca

Generate the title when possible. 
We identified a bug that prevented the section title from displaying when the section was not included in the navigation.
To resolve this issue, we will generate the title whenever possible.
Cherry-pick from e120371

Update sectionTarget parameter to use more descriptive values in completion calculations
Cherry-pick from 80776b0

### DIFF
--- a/.github/workflows/evolution.yml
+++ b/.github/workflows/evolution.yml
@@ -5,9 +5,9 @@ name: CI
 
 on:
   push:
-    branches: [ main, od_2023 ]
+    branches: [ main, od_2023, v0.3.0 ]
   pull_request:
-    branches: [ main, od_2023 ]
+    branches: [ main, od_2023, v0.3.0 ]
 
 jobs:
   build-and-test:
--- a/example/demo_survey/config.js
+++ b/example/demo_survey/config.js
@@ -87,4 +87,5 @@ module.exports = {
   defaultLocale: "fr",
   timezone: 'America/Montreal',
   region: 'CA', // Used for Google Maps localization. See https://developers.google.com/maps/coverage for possible region codes
+  hasSectionProgressBar: true, // If true, show a progress bar at the top of each section
 };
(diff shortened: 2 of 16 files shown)


greenscientist (approved):
Mostly only reviewed commit message, since it's cherry-picks. 

Only one issue ... On the commit "Run test on 0.3.0 branch "
You add "Add the VisitedPlace type and _cloneDeep function in this commit, to fix compilation problems." These should probably have been added to another commits, as just enabling the workflow should not have cause a compile problem. 

But not important for an old stable branch

PR #917, opened 2025-04-02
Fix bug with SectionProgressBar

samuel-duhaime:
# Pull request

## Description
Fix bug with SectionProgressBar
The section completion percentage will now start at 0% when the first section begins and will reach 100% once the last section is initiated.
To achieve this, we will disregard completed sections when calculating the total number of sections.
Test the completed page and the section progress bar in the demo survey.

Move the circular progress at the end.
Move the circular progress at the end of the div so the user will read the title more easily and only focus on the completion rate if needed.  

Generate the title when possible.
We identified a bug that prevented the section title from displaying when the section was not included in the navigation.
To resolve this issue, we will generate the title whenever possible.

Update sectionTarget parameter to use more descriptive values in completion calculations

### DIFF
--- a/packages/evolution-common/src/utils/helpers.ts
+++ b/packages/evolution-common/src/utils/helpers.ts
@@ -640,7 +640,7 @@ export const isSectionComplete = ({
  * @param {UserInterviewAttributes} options.interview - The interview object.
  * @param {Object} options.sections - The sections object.
  * @param {string} options.sectionName - The shortname of the current section.
- * @param {string} options.sectionTarget - The target section ('current' or 'next').
+ * @param {string} options.sectionTarget - The target section for the completion rate ('currentSection' or 'nextSection').
  * @returns {number} - The calculated completion percentage.
  */
 export const calculateSurveyCompletionPercentage = ({
@@ -652,28 +652,31 @@ export const calculateSurveyCompletionPercentage = ({
     interview: UserInterviewAttributes;
     sections: { [key: string]: unknown };
     sectionName: string;
-    sectionTarget: 'current' | 'next';
+    sectionTarget: 'currentSection' | 'nextSection';
 }): number => {
-    const LAST_SECTION_NAME = 'completed';
+    const LAST_SECTION_NAME = 'completed'; // NOTE: Only survey with the last section named 'completed' are supported for this function.
     const MINIMUM_COMPLETION_PERCENTAGE = 0;
     const MAXIMUM_COMPLETION_PERCENTAGE = 100;
 
+    // Get the stored completion percentage from the interview responses
     const storedCompletionPercentage: number =
         interview?.responses?._completionPercentage || MINIMUM_COMPLETION_PERCENTAGE;
+
+    // Count the number of sections without the last one
     const sectionsWithoutCompleted = Object.keys(sections).filter((sectionName) => sectionName !== LAST_SECTION_NAME);
     const totalSectionsWithoutCompleted = sectionsWithoutCompleted.length;
-    let targetSectionIndex = Object.keys(sections).findIndex((key) => key === sectionName);
 
-    // Increment the index if the section target is 'next'
-    if (sectionTarget === 'next') {
-        targetSectionIndex = Math.min(targetSectionIndex + 1, totalSectionsWithoutCompleted);
+    // Increment the index if the section target is 'nextSection'
+    let sectionTargetIndex = Object.keys(sections).findIndex((key) => key === sectionName);
+    if (sectionTarget === 'nextSection') {
+        sectionTargetIndex = Math.min(sectionTargetIndex + 1, totalSectionsWithoutCompleted);
     }
 
-    const currentCompletionPercentage = Number(((targetSectionIndex / totalSectionsWithoutCompleted) * 100).toFixed(0));
+    // Return the maximum survey completion percentage
+    const targetCompletionPercentage = Number(((sectionTargetIndex / totalSectionsWithoutCompleted) * 100).toFixed(0));
     const completionPercentage = Math.min(
         MAXIMUM_COMPLETION_PERCENTAGE,
-        Math.max(storedCompletionPercentage, currentCompletionPercentage)
+        Math.max(storedCompletionPercentage, targetCompletionPercentage)
     );
-
     return completionPercentage;
 };
--- a/packages/evolution-common/src/utils/helpers.ts
+++ b/packages/evolution-common/src/utils/helpers.ts
@@ -631,3 +631,49 @@ export const isSectionComplete = ({
 
     return isSectionComplete;
 };
+
+/**
+ * Calculate the survey completion percentage based on the number of completed sections.
+ * The percentage begins at 0% when starting the first section and reaches 100% when the last section is initiated.
+ *
+ * @param {Object} options - The options object.
+ * @param {UserInterviewAttributes} options.interview - The interview object.
+ * @param {Object} options.sections - The sections object.
+ * @param {string} options.sectionName - The shortname of the current section.
+ * @param {string} options.sectionTarget - The target section ('current' or 'next').
+ * @returns {number} - The calculated completion percentage.
+ */
+export const calculateSurveyCompletionPercentage = ({
+    interview,
+    sections,
+    sectionName,
+    sectionTarget
+}: {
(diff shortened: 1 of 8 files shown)


packages/evolution-frontend/src/components/pageParts/SectionProgressBar.tsx
```diff
@@ -16,13 +16,23 @@ type SectionProgressBarProps = {
...
-    // Calculate the completion percentage based on the current section index and total sections.
-    // The percentage will be 100% when the last section (completed section) is started.
-    const maximumCompletionPercentage: number = interview?.responses?._completionPercentage || 0;
-    const currentSectionIndex = Object.keys(sections).findIndex((key) => key === shortname) + 1;
-    const totalSections = Object.keys(sections).length;
-    const currentCompletionPercentage = Number(((currentSectionIndex / totalSections) * 100).toFixed(0));
-    const completionPercentage = Math.max(maximumCompletionPercentage, currentCompletionPercentage);
+    // Calculate the completion percentage based on the current section index and total number of sections.
+    // The percentage begins at 0% when starting the first section and reaches 100% when the last section is initiated.
+    const LAST_SECTION_NAME = 'completed';
```
greenscientist, line 21: Is it possible for people to have a last section that is not called completed?

packages/evolution-frontend/src/components/pageParts/SectionProgressBar.tsx
```diff
@@ -16,13 +16,23 @@ type SectionProgressBarProps = {
...
+    const MINIMUM_COMPLETION_PERCENTAGE = 0;
+    const MAXIMUM_COMPLETION_PERCENTAGE = 100;
+    const storedCompletionPercentage: number =
```
greenscientist, line 24: Kind of wonder, should we store the completion percentation or compute it everytime?

packages/evolution-frontend/src/services/display/frontendHelper.ts
```diff
@@ -150,16 +150,24 @@ export const validateButtonActionWithCompleteSection: ButtonAction = (
             if (typeof saveCallback === 'function') {
                 saveCallback(callbacks, updatedInterview, path);
             } else {
-                // Calculate the completion percentage based on the next section index and total sections.
-                // The percentage will be 100% when the last section (completed section) is started.
+                // Calculate the completion percentage based on the next section index and total number of sections.
```
greenscientist, line 153: This code seems quite similar to the one in SectionProgressBar. You also repeast the constant. Should this be put in a common function??

packages/evolution-generator/src/scripts/generate_section_configs.py
```diff
@@ -151,11 +151,12 @@ def generate_section_code(previousSection, nextSection):
                 ts_section_code += f"{INDENT}nextSection: nextSectionName,\n"
                 if parent_section is not None:
                     ts_section_code += f"{INDENT}parentSection: parentSectionName,\n"
-                if title_en and title_fr is not None and in_nav == True:
+                if title_en and title_fr is not None:
```
greenscientist, line 154: You should probably put this change in a separate commit, it's not the same thing.  (Can be the same PR, but I think it fixses something different that the rest. That would help future bisect)

greenscientist (changes requested):

packages/evolution-frontend/src/components/pageParts/SectionProgressBar.tsx
samuel-duhaime: @greenscientist  Currently, we store the completion percentage after each section. This could be very useful information for paradata analysts.

packages/evolution-frontend/src/services/display/frontendHelper.ts
samuel-duhaime: Yes, good idea.

samuel-duhaime:
The circular progress has now moved to the end of the div. It was weird to have it at the start (the user was focusing too much attention on it).

image

packages/evolution-generator/src/scripts/generate_section_configs.py
samuel-duhaime: Done.

greenscientist:
@samuel-duhaime so when people go back in the survey we don't want it to go back to a lower percentage?

samuel-duhaime:
> @samuel-duhaime so when people go back in the survey we don't want it to go back to a lower percentage?

@greenscientist No, we don't want to go back to a lower percentage, except for a few rare cases that are hard to identify with the correct Evolution logic (for example, responding to a question that changes all the survey), the survey completion rate are the same. 

It would be discouraging for the user to be at only 0% if they return to the first section.

packages/evolution-common/src/utils/helpers.ts
```diff
@@ -631,3 +631,49 @@ export const isSectionComplete = ({
...
+    sections,
+    sectionName,
+    sectionTarget
+}: {
+    interview: UserInterviewAttributes;
+    sections: { [key: string]: unknown };
+    sectionName: string;
+    sectionTarget: 'current' | 'next';
+}): number => {
+    const LAST_SECTION_NAME = 'completed';
```
greenscientist, line 657: My only last concern is this constant here. To ensure that people use the right name, we should probably export it and use it elsewhere. (LIke the generator should ensure to add this last section with the right name.)

packages/evolution-common/src/utils/helpers.ts
```diff
@@ -631,3 +631,49 @@ export const isSectionComplete = ({
...
+    const MAXIMUM_COMPLETION_PERCENTAGE = 100;
+
+    const storedCompletionPercentage: number =
+        interview?.responses?._completionPercentage || MINIMUM_COMPLETION_PERCENTAGE;
+    const sectionsWithoutCompleted = Object.keys(sections).filter((sectionName) => sectionName !== LAST_SECTION_NAME);
+    const totalSectionsWithoutCompleted = sectionsWithoutCompleted.length;
+    let targetSectionIndex = Object.keys(sections).findIndex((key) => key === sectionName);
+
+    // Increment the index if the section target is 'next'
+    if (sectionTarget === 'next') {
```
greenscientist, line 668: 'next' should probably also be in some kind of global constant no??

packages/evolution-common/src/utils/helpers.ts
```diff
@@ -631,3 +631,49 @@ export const isSectionComplete = ({
...
+    sectionName: string;
+    sectionTarget: 'current' | 'next';
+}): number => {
+    const LAST_SECTION_NAME = 'completed';
+    const MINIMUM_COMPLETION_PERCENTAGE = 0;
+    const MAXIMUM_COMPLETION_PERCENTAGE = 100;
+
+    const storedCompletionPercentage: number =
+        interview?.responses?._completionPercentage || MINIMUM_COMPLETION_PERCENTAGE;
+    const sectionsWithoutCompleted = Object.keys(sections).filter((sectionName) => sectionName !== LAST_SECTION_NAME);
```
greenscientist, line 663: Can you add a comment here that explain that you count the number of sections. (And explain why to filter out of the last one named completed.

greenscientist (approved):
I'll approve, but please add the requested comment. 

Also, do what you want with my comment about the section name constant. Feel free to add a //TODO or create an issue if you fell this should be addressed later.

packages/evolution-frontend/src/components/pageParts/Section.tsx
```diff
@@ -108,7 +108,7 @@ export const Section: React.FC<SectionProps & WithSurveyContextProps> = (
                         <SectionProgressBar
                             title={props.sectionConfig.title}
                             interview={props.interview}
-                            shortname={props.shortname}
+                            sectionName={props.shortname}
```
GabrielBruno24, line 111: Should shortname be changed to sectionName in this file's props as well?

GabrielBruno24 (approved):

packages/evolution-frontend/src/components/pageParts/Section.tsx
samuel-duhaime: @GabrielBruno24  No shortname has been used in many Evolution files. That would be a massive refactoring that I'm not ready to do. Let's keep this like this.

Issue #858, opened 2025-02-14
survey: move server-side all conditional and validation checks (?)


tahini:
In a pass to cleanup and properly document redux actions (and what happens during their execution), it appears that there's a bit of gymnastic required because we have some values that are updated by the server (the `serverFieldUpdate`) and also validations that need to be done server side (like access code validation). Mechanisms were developed to support those and required a bit of changes client side, because at the time, we wanted to avoid breaking something that worked fine.

Now that we are breaking everything anyway, how about we move all interview side-effect code, like validations and conditional checks, to the server. Anyway, whenever these are called in the `startUpdateInterview` there's a call to the server, except in the case where there are no changes, like when submitting a section, but typically, a callback is called at that moment and that callback makes a server call. 

What this implies is 

1. that a good part of what is currently in `evolution-frontend/src/actions/Survey[Admin].ts` and the files in the folder `evolution-frontend/src/actions/utils` would need to go to the backend. Widget preparation would be done there. The server would return a `RuntimeInterviewAttributes` object ready to be displayed.
2. The widget config could be split (by webpack or the compiler), so that some fields go to frontend (labels), others to backend (conditional, validations) others to both (type, inputType, etc)
3. Frontend is a lot dumber
4. Backend has more work to do
5. Would network latency be any more issue than it is currently? Probably not since there's one request anyway and pretty much the same code will need to run.
6. Would the server resource requirement be increased?

Thoughts?

samuel-duhaime:
@tahini  What is the main advantage of doing that?

tahini:
Advantages: 

* No more distinctions and different code paths for client and server conditionals/side effects (aka serverFieldUpdate) and validations.
* The server would be the only one to have the truth on the content of the interview: he receives updates to responses, updates the interview accordingly and sends back the changes to the client, who simply updates its own version of the interview, unlike now where the client interview changes, the server may change it too and chances are they may be discrepancies.
* Client is more lightweight
* Fwiw, we'd be a step closer to server-side rendering, that you,  @samuel-duhaime, defended the other day ;-)
* There may not be need for multiple back and forth when updating response, changing section, preloading section, getting the widgets of the new section. Currently, changing section will trigger 2 or 3 server updates, sometimes more.
* We already depend on the network at each response change, so we don't really have an "offline" mode anyway.

Disadvantage:

[...]

Issue #785, opened 2024-11-15
survey: timestamps should all be generated at the same place


tahini:
The server should be responsible to timestamp event (start, completion, updates, section events, etc). Also, individual surveys should not have to worry about that.

Currently, it's a bit arbitrary where those timestamps come from. Mostly from the browsers, in the preload or saveCallback functions of the individual survey, so a bit error-prone.

Having a single autority for the clock will allow times differences to be meaningful and comparable.

For now, only the timestamp in the `logs` field all come from the server and can be used for comparison. And in most surveys, the timestamps in the _sections data come from the browser and should also be comparable.

Issue #742, opened 2024-10-18
Streamline section changes (questionnaire navigation) and how they are logged


tahini:
Currently, when clicking on a section's `submit` button, the button widget needs to call an action that will automatically validate the current section and navigate to the next section as per the section's configuration. That next section will thus be considered as started.

But many surveys have the kind of workflow where we loop through a few sections for each household member, before actually reaching the last sections. In this case, in this the last section's preload function that needs to tell the survey to change the active section, by setting the `_activeSection` value to where it needs to go.

This puts the responsibilty of the section navigation to individual survey's section preload, and it also has the consequence of setting the last section as started, even if the participant was never in this section.

The section navigation in the survey should be handled by some API who should figure out what the next section is and the logs should only consider a section started if the participant actually saw the page.

tahini:
This will be a major breaking change for current surveys, but it needs to be done. Should we tag a branch `legacy` (or `v.0.2.2` as we currently have) for our current surveys where we can backport compatible changes and make a totally divergent `main` branch? And maybe change that famous 0.2.2 version that we have in all our package.json files?

greenscientist:
Yeah, we probably abuse a bit too much the term "legacy" in our projects. But tagging (in branching if we need it) is probably a good option. 

Not fully sure what's the best version scheme.   Maybe we can use the opportunity to bump the current version to 0.3.  When you do your breaking change, we can bump it to 0.5 (Yeah, I ski the .4, I want to mark the break!).   After that, we should take the habit of increasing the number (0.6, 0.7, 0.8) for each breaking change like this one. No need to version everything thing, but that will give a quick idea that a specific version of the survey will need major rework based on it's version number. This would me mostly informative, as the actual version tracking is done with the submodule hash version.

tahini:
I foresee there will be a truck load of breaking changes in the upcoming weeks/months, so we might as well branch a `v0.3` version (for the sake of the currently running surveys), then, when we do the first major breaking change (so badly breaking that it would be a mess to upgrade current surveys), we'll upgrade `main` to 0.5 and we say we don't use it in production until it's "ready".

Does that sound like a plan?

Other option is to keep `main` production ready, and I do the proposed changes in a `v0.5` branch. Ultimately, the result is the same, it's just the status of the `main` branch that changes.

greenscientist:
[...]

Issue #523, opened 2024-06-06
Redesign interview logs


kaligrafy:
Create a log table and class. This class should record, for ecah change in the interview data:
- device
- user agent/browser data
- ip
- timestamp
- interview language
- objectName (interivew, household, person...)
- attribute (attribute path in the object: example: age) (undefined if action = create or action = delete an entire object)
- action (create, update, delete)
- value (new value for update, undefined for delete)
Any input on this? @tahini @greenscientist @samuel-duhaime

tahini:
Related: #47

We should investigate external tools that could be used to do all this

tahini:
Now that I'm reading all the related issues and started work on sending the actual user action to the backend, I'm wondering just how complex or simple this needs to be.

If we go for specific fields (like object, path, value, etc), then there are many disjoint cases to cover (user clicked on button, user changed an answer, there was side-effect or none, user got an error, etc). We would then have many `nulll` fields of every row, or else, have different tables, which complexify action handling.

Another option, would be just dump a json string in a text field, along with interview id, timestamp, user_id (null for participant's own) and action type in separate fields.

Advantage of this last approach would be that 

* we don't need to know in advance all the paradata that we will track
* it's easy and quick to insert, no processing of the values to do (except to convert as string, which anyway needs to be done to send json data to the db)
* one record per request
* we can still filter on event type if required to return only desired paradata

Disadvantage is that it is harder to query, but then we can build materialized view on demand and the paradata analysis is not something that is done often or is time-sensitive.

Thoughts? @kaligrafy @samuel-duhaime  

@greenscientist I think this quite resembles the structured login structure that you've been proposing in other issues, but saved in our DB, close to the data and without external tool

samuel-duhaime:
@tahini, I don't have much experience redesigning the interview logs, but should we also save the platform ('survey' or 'admin')? 

@greenscientist, I'm curious about your opinion on this issue.

tahini:
@samuel-duhaime for the platform, survey is when user_id is null (as it is a participant and already linked with the interview) and admin is cases with a user_id

PR #935, opened 2025-05-01
survey update: add field to differentiate the user action from side e…

### LINKED ISSUE #814: logs: differentiate values set by participant from those set by the survey
The `updateInterview` path receives an object with the valuesByPath modified from the client. But these values can come either from the participant input or as a side-effect of calculations from current interview state.

These 2 kinds of values should be differentiated so that log analysis can treat them differently depending of their provenance.

tahini:
…ffects

fixes #814

Add a `userAction` field in the `startUpdateInterview` callback data parameters.

For now, this user action parameter can either be a `buttonClick` with the ID of the button clicked by the user, or a `widgetInteraction` with a widget type and the path and value of the widget clicked. For the widget interacted with, this path and value replaces what was previously in the `valuesByPath` object, which now only contains the side-effects of the user interaction.

The `Button` and `Question` widgets now provide this user interaction when calling the `startUpdateInterview` callback. And this value is sent to the backend in the `updateInterview` and the
`updateValidateInterview` calls.

In the backend, the `userAction` parameter is received and we make sure that a valid request has either a `valuesByPath` or `userAction` parameter. For now, the interview update simply merges the widget interaction user action data to the valuesByPath to avoid having to change this code path right now, as it will be revisited (and simplified) soon when side effects are in the backend.

### DIFF
--- a/packages/evolution-backend/src/services/interviews/interviewUtils.ts
+++ b/packages/evolution-backend/src/services/interviews/interviewUtils.ts
@@ -4,27 +4,39 @@
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
  */
-/** Change the prefixes in the valuesByPath and unsetPaths from responses to validated_data. */
+
+import { UserAction } from 'evolution-common/lib/services/questionnaire/types';
+
+/** Change the prefixes in the valuesByPath, unsetPaths and userAction from
+ * responses to validated_data. */
 export const mapResponsesToValidatedData = (
     valuesByPath: { [key: string]: unknown },
-    origUnsetPaths: string[]
-): { valuesByPath: { [key: string]: unknown }; unsetPaths: string[] } => {
+    unsetPaths: string[],
+    userAction?: UserAction
+): { valuesByPath: { [key: string]: unknown }; unsetPaths: string[]; userAction?: UserAction } => {
     // Content in valuesByPath and unsetPaths with prefix 'responses' should
-    // be put in 'validated_data' instead. This route should not modify the
-    // responses
+    // be put in 'validated_data' instead.
     const renameKey = (oldKey: string): string => {
         if (oldKey.startsWith('responses.') || oldKey === 'responses') {
             return oldKey.replace('responses', 'validated_data');
         }
         return oldKey;
     };
+
+    // Map the values by path keys
+    const updatedValuesByPath = {};
     Object.keys(valuesByPath).forEach((key) => {
         const newKey = renameKey(key);
-        if (newKey !== key) {
-            valuesByPath[newKey] = valuesByPath[key];
-            delete valuesByPath[key];
-        }
+        updatedValuesByPath[newKey] = valuesByPath[key];
     });
-    const unsetPaths = origUnsetPaths.map(renameKey);
-    return { valuesByPath, unsetPaths };
+    // Map the unset paths
+    const updatedUnsetPaths = unsetPaths.map(renameKey);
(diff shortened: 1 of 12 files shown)


tahini:
This is a first step before cleaning up the paradata logging a little bit more (that will be in the backend). Work on #523, #785, #742, #47 and #858 will follow and this will hopefully simplify the backend update code a bit.

kaligrafy (approved):
Les commentaires de Copilot semblent legit, il faudrait vérifier

packages/evolution-backend/src/services/interviews/interviewUtils.ts
```diff
@@ -4,11 +4,15 @@
...
+import { UserAction } from 'evolution-common/lib/services/questionnaire/types';
+
 /** Change the prefixes in the valuesByPath and unsetPaths from responses to validated_data. */
 export const mapResponsesToValidatedData = (
     valuesByPath: { [key: string]: unknown },
-    origUnsetPaths: string[]
-): { valuesByPath: { [key: string]: unknown }; unsetPaths: string[] } => {
+    origUnsetPaths: string[],
+    userAction?: UserAction
+): { valuesByPath: { [key: string]: unknown }; unsetPaths: string[]; userAction?: UserAction } => {
```
samuel-duhaime, line 15: unsetPaths or origUnsetPaths ?

packages/evolution-frontend/src/actions/Survey.ts
```diff
@@ -88,15 +88,53 @@ export const updateSection = (
     return [interview, currentValuesByPath];
 };
 
+export const updateInterviewData = (
```
samuel-duhaime, line 91: Can you add a JSDoc for this one?

packages/evolution-frontend/src/components/survey/Question.tsx
```diff
@@ -101,31 +102,49 @@ export class Question extends React.Component<QuestionProps & WithSurveyContextP
...
+            if (userAction === undefined) {
+                userAction = {
+                    type: 'widgetInteraction',
+                    widgetType: 'string',
+                    path: 'responses.' + this.props.customPath,
+                    value: parsedCustomValue
+                };
+            } else {
+                console.warn(
+                    'userAction already defined, but a custom value has been specified. It is thus possible to update more than one value for a single user action.'
```
samuel-duhaime, line 132: Typo for 'It is thus possible'?

samuel-duhaime (changes requested):

packages/evolution-backend/src/services/interviews/interviewUtils.ts
tahini: La valeur de retour c'est bien unsetPaths

packages/evolution-frontend/src/components/survey/Question.tsx
tahini: > Typo for 'It is thus possible'?

Non, je ne crois pas. En fait, c'est un warning qu'on ne devrait pas voir, mais qu'on met justement au cas où.

samuel-duhaime (approved):
J'accepte, il reste seulement le commentaire pour le JSDoc.

packages/evolution-frontend/src/actions/Survey.ts
tahini: I did, and there will be a big warning in caps to not call this function ever from other files than the redux actions! It unfortunately needed to be exported so the SurveyAdmin.ts file can call it, but really, it is not part of the API.

PR #939, opened 2025-05-06
Add home section demo generator

samuel-duhaime:
# Pull request

## Description
Add home section demo generator
We need to add a second introduction to be able to have a footer in the landing page.
The survey.yml is a generic text with no specific survey information.
The completeLandingPageTests need to start in French.


samuel-duhaime:
Close for now, it was harder to rebase than to start a new PR.

PR #964, opened 2025-05-09
Paradata: redesign interview logs

### LINKED ISSUE #523: Redesign interview logs

tahini:

fixes #523 

* Add a table for paradata events (deletes current `logs` column after having saved the data in the new table as legacy events, rollback migration has been tested and works to bring back the log statements in interview table)
* Update log exports accordingly
* Update the paradata events logging to save with the new paradata events format (server, side_effect, button click or widget interaction)


kaligrafy (approved):

packages/evolution-backend/src/models/migrations/20250429094700_createParadataTable.ts
```diff
@@ -0,0 +1,145 @@
...
+    await knex.schema.createTable(paradataTbl, (table: Knex.TableBuilder) => {
+        table.integer('interview_id').notNullable();
+        table.foreign('interview_id').references(`${interviewsTbl}.id`).onDelete('CASCADE');
+        table.timestamp('timestamp', { useTz: true }).notNullable().defaultTo(knex.fn.now());
+        // Value will be null if user is the participant, the ID of the user otherwise, that can't be deleted
+        table.integer('user_id').references('users.id').onDelete('RESTRICT');
+        table
+            .enu(
+                'event_type',
+                [
```
samuel-duhaime, line 79: Should we have an event_type for help pop-up?

packages/evolution-backend/src/models/migrations/20250429094700_createParadataTable.ts
```diff
@@ -0,0 +1,145 @@
...
+        for (let j = 0; j < paradata.length; j++) {
+            const logEntry = paradata[j];
+            const { timestamp, event_type, event_data } = logEntry;
+            const { valuesByPath, unsetPath } = event_data;
+
+            const log: ParadataLog = {
+                timestamp: Math.floor(new Date(timestamp).getTime() / 1000), // Convert to Unix seconds
+                valuesByPath
+            };
+            if (event_type === 'legacy_server' || event_type === 'server_event') {
```
samuel-duhaime, line 131: Does event_type have a type EventType or LogsEventType ?

packages/evolution-backend/src/models/paradataEvents.db.queries.ts
```diff
@@ -0,0 +1,41 @@
...
+
+const log = async ({
+    interviewId,
+    userId,
+    eventType,
+    eventData
+}: {
+    interviewId: number;
+    userId?: number;
+    eventType: string;
```
samuel-duhaime: AS written before, I think we should type this eventType, so we don't have 1000 event types.

packages/evolution-backend/src/models/migrations/20250429094700_createParadataTable.ts
tahini: > Should we have an event_type for help pop-up?

We'll add event types as we actually support them, but indeed, it would make sense to add it to this paradata here as a user action.

packages/evolution-backend/src/models/migrations/20250429094700_createParadataTable.ts
tahini: > Does event_type have a type EventType or LogsEventType ?

What do you mean here? Technically, its type is the `ParadataEventType` defined in the `paradataEvents.db.queries.ts` file. But since this file is a migration we can't directly call the function, as if we call them, especially the `down` part, it's that we are playing between versions of the code and the function we want to call may not be available.

packages/evolution-backend/src/models/paradataEvents.db.queries.ts
tahini: Je pense que tu as review commit par commit car le 3e commit ajoute le type, quand on supporte les nouvelles paradata. Je peux mettre ce changement dans le premier commit au besoin, ça n'avait pas besoin d'attendre

samuel-duhaime: Ah, non c'est bon !

samuel-duhaime (approved):

PR #988, opened 2025-05-16
Fix poetry unit test and add to worflows

samuel-duhaime:
# Pull request

## Description
Fix Python tests for Evolution-Generator 
Update the path of all the tests from an old structure version to the new structure.
Fix all the tests, some of the error messages have been changed since the last time we tested.
Ignore some of the generated output and input of these tests.
Install pytest package in Poetry.

Run Python unit tests inside the Evolution-Generator worflows
I did add some .gitignore file because we need an empty folder to save the generated files from the tests.

### DIFF
--- a/.github/workflows/evolution-generator.yml
+++ b/.github/workflows/evolution-generator.yml
@@ -29,6 +29,10 @@ jobs:
     - name: Run Poetry
       working-directory: packages/evolution-generator
       run: poetry install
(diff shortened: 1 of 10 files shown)


samuel-duhaime:
I will fix this next week, it failed, because the folders aren't saved. Probably need to make a small changes to the .gitignores.

tahini (approved):

PR #990, opened 2025-05-20
ci: add sequential tests to the CI

tahini:
Using a postgis database, this executes Evolution's sequential test for every pull request.


samuel-duhaime (changes requested):
I'm waiting for all the tests to pass.

samuel-duhaime (approved):

PR #1006, opened 2025-05-27
survey: extract a `Survey` component for both admin and participant

tahini:
`Survey` and `SurveyCorrection` differed only by the values of their actions and the signature of the startSetInterview redux action.

The common part of both components is thus extracted to a `Survey` pageParts component and receives the redux action callbacks as props. The interview is set in the wrapping components and is not required anymore, as it was only called in the `componentDidMount` method.

Two components `SurveyCorrection` and `SurveyParticipant` set the redux actions and load the interview respectively for the correction and participants modes. This results in much smaller components and reduces code duplication between survey components.

All components are function components.


packages/evolution-frontend/src/components/admin/pages/SurveyCorrection.tsx
```diff
@@ -7,283 +7,85 @@
...
-            )
-        );
-    const startRemoveGroupedObjectsAction: StartRemoveGroupedObjects = (paths, callback, returnOnly) =>
-        dispatch(startSurveyCorrectedRemoveGroupedObjects(paths, callback, returnOnly));
+    React.useEffect(() => {
+        // Load the interview when first mounting the component
+        dispatch(startSetSurveyCorrectedInterview(interviewUuid));
+    }, [interviewUuid, dispatch]); // Reload the interview if the uuid changes
+
+    if (!interviewLoaded || !interview || !interview.sectionLoaded) {
```
samuel-duhaime, line 76: For the LoadingPage. It might be better to use Suspense (not totally sure).

```
<Suspense fallback={<Loading />}>
  <SomeComponent />
</Suspense>
```

Source: https://react.dev/reference/react/Suspense

packages/evolution-frontend/src/components/pageParts/Survey.tsx
```diff
@@ -0,0 +1,203 @@
...
+        props.surveyContext.sections[activeSection].customStyle
+    ) {
+        customStyle = props.surveyContext.sections[activeSection].customStyle as React.CSSProperties;
+    }
+
+    let surveyContainerStyle: React.CSSProperties = {};
+    if (config.logoPaths) {
+        surveyContainerStyle = {
+            backgroundImage: `url(${config.logoPaths[i18n.language]}), url(/dist/images/ornaments/ornament_flower_points_pale.svg)`,
+            backgroundSize: '15rem, 6rem',
```
samuel-duhaime, line 143: Can we put that in a CSS class ?

packages/evolution-frontend/src/components/pages/SurveyParticipant.tsx
```diff
@@ -0,0 +1,100 @@
...
+            surveyUuid,
+            interviewContext &&
+                interviewContext.status === 'entering' &&
+                Object.keys(interviewContext.response).length > 0
+                ? interviewContext.response
+                : undefined
+        );
+    }, [surveyUuid, dispatch]); // Re-run when survey uuid changes
+
+    if (!interviewLoaded || !interview || !interview.sectionLoaded) {
```
samuel-duhaime, line 85: Maybe Suspense also.

samuel-duhaime (changes requested):

packages/evolution-frontend/src/components/pageParts/Survey.tsx
```diff
@@ -0,0 +1,203 @@
...
+            backgroundPosition: 'left 1rem bottom 1rem, right 1rem top 100.5%',
+            backgroundRepeat: 'no-repeat, no-repeat'
+        };
+    }
+
+    return (
+        <div className="survey" style={surveyContainerStyle}>
+            <div style={{ width: '100%', margin: '0 auto' }}>
```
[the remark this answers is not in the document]
tahini: > @tahini Peux-tu l'enlever à la place. Aucune raison de maintenir un CSS qui ne fait rien.

Parce que _je crois_ qu'il ne fait rien ne veut pas dire qu'il ne fait rien. Moi, j'y connais rien, peut-être qu'il a un effet dans internet explorer 6 ou sur safari... j'aimerais mieux laisser une autorité plus compétente en juger...

PR #1034, opened 2025-06-13
Fix demo generator

tahini:


samuel-duhaime (approved):

samuel-duhaime:
Make sure to remove the double click that I had to insert in the next button test in https://github.com/chairemobilite/evolution/pull/1010, so we can test this.

PR #1037, opened 2025-06-16
generator: add household section and improved context-aware label gen…

### LINKED ISSUE #440: helpers: Add person and gendered suffixes helpers in mainline
Many surveys define specific helpers to retrieve the person involved in the question (with nickname), the person count, as well as the gendered suffixes for label display.

For example od_mtl_2023 and od_longuedistance_2023 both have these functions in helper files. The generator now needs them to properly generate the labels, so it imports something that is expected to exist at a specific location in the survey. These concept are basic enough that they should be moved to Evolution itself. It will also force us to streamline our handling of person and household data.

### LINKED ISSUE #606: generator: Add the possibility to add extra fields to labels
Currently, the generator only accepts simple labels. If we want to add variables to the labels, eg. "Here follow questions about your trip to {{destination}}", we need to use a custom widget.

2 possible solutions (not mutually exclusive):

1. Just support customLabels: like customConditionals and customValidations, these can be written by survey developers and be as complicated as necessary (and we do have very complicated labels in some surveys).
2. Support variables in normal labels: we already have special cases for labels to integrate gendered suffix labels and persons count labels. We could add a column containing the paths of the responses to add, document that the labels can contain those named `{{path after last dot}}` and they can be added to the `t` function call.

Solution 1 is relatively straightforward given the other custom stuff we have. 
@samuel-duhaime any idea how we could implement solution 2 to support multiple variables from the excel file?

### LINKED ISSUE #956: demo_generator: Add the householdMembers section
# Issue

## Description
Use the Evolution-Generator for this. Conduct unit tests and UI tests. Take the same questions as od_nationale_2024.

tahini:
…eration

fixes #956, fixes #440, fixes #606

Add household section to Demo_Generator with support for 1 and 2-person households. Introduce countAdults helper
BREAKING CHANGE: Rename all preload to customPreload in all surveys. Update tasks.json to allow running Demo_Generator in dev mode.

Translation & Label Improvements:
Add count, nickname, and gender contexts for translations using the Generator. Add getActivePersonId, getActivePersonGender, and getCurrentPersonId helpers. Generate dynamic labels using templates like {{count}}, {{nickname}}, and gender-specific suffixes. Add unit tests for context helpers and update label generation tests accordingly. BREAKING CHANGES: Rename fr/en columns in Widgets and Choices sheet to label::fr and label::en.

Household UI & Widgets:
Add all household widgets in Demo_Generator.
Add 2 custom map widgets with geojson support.

Gender Handling:
Generate labels with male, female, and other gender options. Add adultAge to config.

Bug Fixes:
Fix bug where translation labels were duplicated across all locale files. Fix label generation logic and update household Excel accordingly.

Conditionals:
Add logic to use '{{currentPerson}}' in the path.

Documentation:
Update README with details on gender context and dynamic label support.


tahini:
@kaligrafy, c'est la PR de Samuel, donc j'ai fixé les derniers commentaires.

PR #1051, opened 2025-06-30
generator: let radioNumber widgets support a field for min/max

### LINKED ISSUE #1050: generator: let RadioNumber support a field for min/max values
The radioNumber can be bounded on the value of a field (for example the number of hybrid or electric cars), so if the max/min value is a string, it should generate a function that returns this response.

tahini:
fixes #1050

If the min/max value entered is not an integer, it is considered a field value and the min/max value will be replaced by a function that gets the corresponding response.

The `generate_widget_statement` function now returns not just the statement, but optional values that hints at additional imports to add to the import statement. This allows to add the `surveyHelper` functions import from the `evolution-common/lib/utils/helpers` file. If the radioNumber widget uses the `surveyHelper` function in its code, the `generate_widget_statement` function will return `True` for `needsHelperImport`.


packages/evolution-generator/src/scripts/generate_widgets.py
```diff
@@ -345,10 +363,18 @@ def generate_import_statements(
...
 ):
     od_survey_helpers_import = ""
     if has_nickname_label or has_gendered_suffix_label or has_persons_count_label:
         od_survey_helpers_import = "import * as odSurveyHelpers from 'evolution-common/lib/services/odSurvey/helpers';\n"
+
+    survey_helper_import = ""
+    if needs_helper_import:
+        survey_helper_import = (
+            "import * as surveyHelper from 'evolution-common/lib/utils/helpers';\n"
+        )
```
GabrielBruno24, line 376: Might be cleaner to use a ternary operator here.

GabrielBruno24 (approved):
Looks good to me

PR #1057, opened 2025-07-02
generator: limit number of splits in path transform

tahini:
If the path had many '.' signs, it would throw an error about too many values to unpack. We thus limit the number of splits, to separate the section and field name.


greenscientist (commented):
Should we have at least a warning here if there's multiple "." ??

tahini:
Avertissement ajouté

Issue #1063, opened 2025-07-03
generator: labels for `infoText` should also add nicknames and other fields


tahini:
See screenshot, this label did not automatically add the code for the nickname for some reason.

![Image](https://github.com/user-attachments/assets/ae733182-9bce-4757-8da3-032d0bd04054)

PR #1070, opened 2025-07-08
Update text property label generation to support dynamic key names with InfoText

samuel-duhaime:
# Pull request
fix: #1063 

## Description
Update text property label generation to support dynamic key names with InfoText

Remove input-text class in the function, because survey-section__text class div is already created. We should, aim for this class if we need to change the style of the InfoText.


tahini (approved):
ça a l'air bon, ça a été testé?

samuel-duhaime:
> ça a l'air bon, ça a été testé?

Oui, testé sur le demo_generator.

PR #1094, opened 2025-07-15
generator: make parameters case-insensitive

### LINKED ISSUE #1083: generator: support the parameters with capital letter at the beginning
In the Excel cell, Libreoffice automatically capitalizes the first letter of a line, so when entering parameters on separate lines, line `min=1`, `max=5`, the `max` becomes `Max` and we have to manually edit it back to small `m`. The generator could be case-insensitive for the parameters, unless there is a good reason to be case sensitive?

tahini:
fixes #1083

Extract a function which parses the parameters field as a key, value dictionary, converting all keys to lowercase.

The `get_parameters_values` function now becomes
`get_radio_number_parameters` to prepare to have more widgets support parameters and allow to get specific parameters per widget.

Update tests for case insensitivity

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_widgets.py
+++ b/packages/evolution-generator/src/scripts/generate_widgets.py
@@ -13,9 +13,9 @@ import re  # Regular expression module for pattern matching
 from typing import TypedDict
 
 
-class ParametersResult(TypedDict):
+class RadioNumberParametersResult(TypedDict):
     """
-    TypedDict to hold the result of parsing parameters from the Widgets sheet.
+    TypedDict to hold the result of parsing parameters for a radio_number widget from the Widgets sheet.
     It contains:
     - min_value: int, the minimum value parsed from parameters.
     - max_value: int, the maximum value parsed from parameters.
@@ -679,7 +679,7 @@ def generate_radio_number_widget(
     - Parses overMaxAllowed from the 'parameters' column (set to true if a line is 'overMaxAllowed').
     - Generates the TypeScript widget code for InputRadioNumberType.
     """
-    parameters: ParametersResult = get_parameters_values(row)
+    parameters: RadioNumberParametersResult = get_radio_number_parameters(row)
     over_max_allowed = parameters["over_max_allowed"]
 
     result: WidgetResult = {"statement": "", "needsHelperImport": False}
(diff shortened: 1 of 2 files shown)


packages/evolution-generator/src/scripts/generate_widgets.py
```diff
@@ -869,56 +869,67 @@ def generate_text_widget(
...
-                result["max_value"] = max_value
-        elif line.strip() == "overMaxAllowed":
+                result["max_value"] = value  # Keep as string if not an integer
+        elif key == "overmaxallowed":
             result["over_max_allowed"] = True
-        elif line.strip() != "":
+        elif key != "":
             print(
-                f"Warning: Unrecognized line in parameters in Widgets sheet: '{line}'. Expected format: min=0\\nmax=6\\noverMaxAllowed."
+                f"Warning: Unrecognized parameter '{key}' for radio_number in Widgets sheet. Expected 'min', 'max' or 'overMaxAllowed'."
```
samuel-duhaime, line 932: Can you show the row number in the warning to help us find it?

packages/evolution-generator/src/tests/test_generate_widgets.py
```diff
@@ -535,95 +536,111 @@ def test_generate_radio_number_widget_unrecognized_parameter_line(capsys):
...
+
+    def test_get_radio_number_parameters_over_max_allowed(self):
+        """Test get_parameters_values parses overMaxAllowed"""
+        row = {"parameters": "min=1\nmax=5\noverMaxAllowed"}
+        params = get_radio_number_parameters(row)
+        assert params["min_value"] == 1
+        assert params["max_value"] == 5
+        assert params["over_max_allowed"] is True
+
+    def test_get_radio_number_parameters_non_numeric_min_max(self, capsys):
```
samuel-duhaime, line 566: Should we skip this warning message when we use a variable (min=abc)?
Or maybe we could do something like that Number(variableName).

samuel-duhaime (changes requested):

packages/evolution-generator/src/tests/test_generate_widgets.py
tahini: The warning just says it cannot compare the min/max, which is ok. It is informative, but normal.

samuel-duhaime: Yeah, but we will have this warning everything we used a variable. Not sure it's what we wanted.

packages/evolution-generator/src/scripts/generate_widgets.py
tahini: I'll add an issue, but no, we don't pass the row id along everything. The widget for row generation should be part of a class, with the row ID and all warnings and errors for this row should be sent to a method that will display the row id automatically.

samuel-duhaime (approved):
C'est bon.

PR #1107, opened 2025-07-21
generator: make the `other` gender translation the default

### LINKED ISSUE #1105: generator: Make sure there is always a fallback translation string
Currently, when there are gendered strings in the labels, for example `{{gender:he/she/they}}`, it creates 3 translation strings with contexts `_male`, `_female` and `_other`. This forces all usages of the context to "manually" make sure they use one of those 3 contexts as there is no fallback string.

There should always be a fallback translation string (one without the context at the end). In our case, the `_other` string should be that one.

tahini:
fixes #1105

If there is no context, or the requested context does not exist, there should always be a fallback translation string to avoid untranslated text. The `other` gender should be this default one.

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_labels.py
+++ b/packages/evolution-generator/src/scripts/generate_labels.py
@@ -303,13 +303,13 @@ def add_translations_from_excel(
                 else None
             )
 
-            # Expand gender context for fr_label and fr_label_one
+            # Expand gender context for labels, by language
             gender_fr = expand_gender(fr_label)
             gender_fr_one = expand_gender(fr_label_one)
             gender_en = expand_gender(en_label)
             gender_en_one = expand_gender(en_label_one)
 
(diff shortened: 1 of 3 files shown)


samuel-duhaime (changes requested):
Petite question. 
Pour un {{gender:a}} ou {{gender:a/b}}, est-ce que ça va bien marcher ?
Je sais que ça marche bien pour {{gender:a/b/c}}

Aussi, je crois qu'il faut modifier la ligne 545 ?

additional_t_context += f"{INDENT}{INDENT}{INDENT}context: personGender === 'male' || personGender === 'female' ? personGender : 'other',\n"

samuel-duhaime (changes requested):
Not sure about context: activePerson?.gender.

What happens if there is no gender specified (for example, a 4-year-old or younger).
What occurs if the gender is not 'male' or 'female'?

We need to be able to handle every case.

samuel-duhaime (approved):

PR #1115, opened 2025-08-11
ui tests: add missing await before nav status check

tahini:
This caused tests to pass, but failure to be reported outside tests.

## Summary by CodeRabbit

- **Tests**
  - Improved UI test stability by ensuring button selection and state assertions fully resolve before proceeding, reducing flakiness and false negatives.
  - Makes navbar-related checks more deterministic across browsers and CI environments.
  - No changes to production code or public interfaces; this only strengthens test reliability and clarity when failures occur.


tahini:
I added the suggestion of coderabbit, except the browser.close, which does not seem to be requried in our case.

PR #1119, opened 2025-08-12
generator: add a fourth option for custom vs other gender string

### LINKED ISSUE #1111: generator: gender strings should have a fourth option for default
Other than `male` and `female`, the gender can be something `custom` that uses another pronouns 'iel' in french, 'they' in english for example. But there should be a forth one, when the gender is simply not set or `preferNotToAnswer` that would use a mix of pronous ('il/elle', 'he/she')

tahini:
fixes #1111

This allows to add another gender option, that will be used as the default, for more neutral tones (like if the participant prefers not to answer), while a custom option is when the gender is some other, but specific, custom option.

The gender string now supports four parts: male/female/custom/other. If there are only 2 parts, custom and other will use the first one (same as male). If there are three parts, other and custom will be the same.

Also add support for quoted gender strings, which allow to have a '/' character inside the quotes. For example
`{{gender:il/elle/iel/"il/elle"}}` would resolve in 4 parts, respectively, "il", "elle", "iel" and "il/elle".

Add tests for all those cases.

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_labels.py
+++ b/packages/evolution-generator/src/scripts/generate_labels.py
@@ -253,6 +253,73 @@ def save_translations(
         raise e
 
 
+def add_gender_or_base_translations(
+    language,
+    section,
+    path,
+    gender_dict,
+    label,
+    extraSuffix,
+    rowNumber,
+    translations_dict,
+):
+    """
+    Adds gender-specific translations to the translations dictionary or simply the label if no gender specific strings.
+
+    Args:
+        language (str): The language code (e.g., 'fr', 'en')
+        section (str): The section name
+        path (str): The base path for the translation key
+        gender_dict (dict or None): Dictionary with gender-specific values or None
+        label (str): The default label value to use if gender_dict is None
+        rowNumber (int): The row number for error reporting
+        translations_dict (dict): The translations dictionary to update
+    """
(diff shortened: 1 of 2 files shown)


packages/evolution-generator/src/scripts/generate_labels.py
```diff
@@ -253,6 +253,73 @@ def save_translations(
...
+            path=path + "_female" + extraSuffix,
+            value=gender_dict["female"],
+            rowNumber=rowNumber,
+            translations=translations_dict[language],
+        )
+        add_translation(
+            language=language,
+            section=section,
+            path=path + "_custom" + extraSuffix,
+            value=gender_dict["custom"],
```
samuel-duhaime, line 299: I'm in reflection, is 'custom' the right word?

samuel-duhaime (approved):
Ça me semble bien, je suis incertain pour le nom 'custom', mais j'ai rien de mieux comme proposition...

packages/evolution-generator/src/scripts/generate_labels.py
tahini: En effet, le 3e choix est 'custom', donc ce sera cette string qui sera utilisé pour ce choix. Le choix "preferNotToAnswer" fera un fallback vers le défaut (other)

PR #1125, opened 2025-08-15
segments: fix icon in the trips and segment template

tahini:
They wrongly used the marker icons, but the icons are not in a map context so it needs to be the icon without marker.

## Summary by CodeRabbit

- **Refactor**
  - Standardized the activity icon reference used in trip and segment visuals for origin, destination, and next destination, improving naming consistency across the interface.
  - No changes to functionality or behavior; icons render as before.
  - This is a non-user-facing update aimed at consistency and maintainability.


tahini:
> Yes, we should make some change in the CSS for the size (and also for the gap between the text and the icon).

_I_ am not going to make any change in the css :p I was just asking if I needed to add any css classes here (and if so, where and what) so that _someone else_ more competent in styling can do it easily after ;-)

kaligrafy:
tu as mis l'icône de garderie pour études dans un autre lieu :-)

tahini:
> tu as mis l'icône de garderie pour études dans un autre lieu :-) 

... j'étais vraiment supposée savoir que `school_building_small.svg` c'était une garderie?

@kaligrafy Mais c'est vrai que j'ai pas attendu ton review pour https://github.com/chairemobilite/evolution/pull/1121 Pourrais-tu faire un review rétrospectif et indiquer quels mappings seraient à changer?

PR #1154, opened 2025-08-22
validations: support Canada or Quebec postal code validations

### LINKED ISSUE #1150: authByField: postal code regex expression should be configurable
And most importantly it should match the one from the frontend widget's.

Currently, in a survey, we have the authByField regex that hard codes a Quebec postal code, while the `postalCodeValidation` used by the generator for the postalCode widget has a canadien postal code regex!

It should be possible to configure this at one place for the whole survey (but backend cannot import widgets file as they require frontend package). In the meantime, configuring the postalCode regex somewhere in the config should be good enough and work for both generator validations and authByField login

tahini:
fixes #1150

This extracts 2 regex for Canada and Quebec postal codes. The survey can be configured to use one of those formats and everywhere it is needed (postal code validation, authentication by access/postal codes) will use this configuration.

Add a `postalCodeRegion` configuration, which can take the value of either `quebec` or `canada`. If undefined, the postal code's content will not be validated with the builtin functions. Surveys can add their own validation functions if need be.

## Summary by CodeRabbit

* **New Features**
  * Region-aware postal code validation (Canada, Quebec, Other) used across surveys and auth flows.
  * Centralized postal-code validator applied in UI and backend.
  * Project configuration adds postalCodeRegion; demo examples set it to Canada.

* **Localization**
  * Added English and French error messages for required/invalid postal codes, including region-specific variants.

* **Tests**
  * Added unit tests and test-only config mocks to verify region-specific behavior.

### DIFF
--- a/example/demo_generator/config.js
+++ b/example/demo_generator/config.js
@@ -65,6 +65,7 @@ module.exports = {
             lng: -72.9215
         }
     ],
+    postalCodeRegion: 'canada',
     detectLanguageFromUrl: true,
     detectLanguage: true,
     languages: ['fr', 'en'],
--- a/example/demo_survey/config.js
+++ b/example/demo_survey/config.js
@@ -55,6 +55,7 @@ module.exports = {
     facebook: false,
     byField: false
   },
+  postalCodeRegion: 'canada',
   logDatabaseUpdates: true,
   askForAccessCode: true,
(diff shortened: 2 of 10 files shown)


kaligrafy (approved):

tahini:
Je confirme que ça fonctionne avec od_nationale

PR #1161, opened 2025-08-25
generator: add gendered suffix labels if gendered already exist

### LINKED ISSUE #1160: generator: add context translation strings for counts even if not present
When translating a key with the `t` function, if a translation key has both `count` and `context` (gender) values, the `context` takes priority over the `count`, which means that, for example,, the translation function `t` will look for keys in this priority order: `key_gender_one`, `key_gender`, `key_one`, and `key`. If the `count` is one, but the _one part does not have gender, it will find they key `key_gender` first and that's what it returns!

To fix this, the generator would need to generate keys for both _one and non-_one variants simultaneously, and ensure that whenever there are gender context keys for the default case, a corresponding key with the same context AND _one suffix also exists.

tahini:
fixes #1160

The translation strings use the gender as context and the plural suffix to address participant directly or with a nickname. In i18n, the context has priority over the plural, so, for example, with both, the labels will be looked for in priority: `label_context_suffix`, `label_context`, `label_suffix`, `label`. In Evolution, the plural (suffix) should have precedence over the context, so we need to make sure that if a label with `label_context` exists and there is plural, then `label_context_suffix` should also exist, otherwise, we'll get the multi-person gendered translation before the single-person one.

## Summary by CodeRabbit

- New Features
  - Translation generation now creates gendered variants when plural/context suffixes are requested, improving localized label coverage.
- Bug Fixes
  - Fixed precedence between plural suffixes and context to produce consistent gendered labels.
  - Propagates existing non-suffixed gender labels to their suffixed counterparts when applicable.
- Tests
  - Added tests validating behavior with extra suffixes and existing gendered labels to ensure correct translation outputs.

### DIFF
--- a/packages/evolution-generator/src/tests/test_generate_labels.py
+++ b/packages/evolution-generator/src/tests/test_generate_labels.py
@@ -448,6 +448,182 @@ class TestAddGenderOrBaseTranslations:
             {},
         )  # Female with suffix
 
+    def test_with_extra_suffix_and_previous_labels(self, monkeypatch):
+        """
+        Test add_gender_or_base_translations with an extra suffix and previous gendered labels.
+        Only one translation string, but all genders should be added anyway as there are already strings with context
+        """
+        # Mock the add_translation function to record calls
+        calls = []
+
+        def mock_add_translation(
+            language, section, path, value, rowNumber, translations
+        ):
+            calls.append((language, section, path, value, rowNumber, translations))
+
+        monkeypatch.setattr(
+            "scripts.generate_labels.add_translation", mock_add_translation
+        )
+
+        # Setup test data
+        language = "en"
+        section = "test_section"
+        path = "test.path"
+        gender_dict = None
+        label = "Test label"
+        extraSuffix = "_one"
+        rowNumber = 1
+        translations_dict = {
+            "en": {
+                "test_section": {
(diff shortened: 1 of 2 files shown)


packages/evolution-generator/src/tests/test_generate_labels.py
```diff
@@ -448,6 +448,182 @@ def mock_add_translation(
             {},
         )  # Female with suffix
 
+    def test_with_extra_suffix_and_previous_labels(self, monkeypatch):
+        """
+        Test add_gender_or_base_translations with an extra suffix and previosu gendered labels.
```
samuel-duhaime, line 453: Small typo: 'and previosu'

packages/evolution-generator/src/tests/test_generate_labels.py
```diff
@@ -448,6 +448,182 @@ def mock_add_translation(
...
+            path,
+            gender_dict,
+            label,
+            extraSuffix,
+            rowNumber,
+            translations_dict,
+        )
+
+        # Verify the calls to add_translation
+        assert len(calls) == 4  # Should have 3 calls (other, male, female, custom)
```
samuel-duhaime, line 503: Typo: // Should have 4 calls

samuel-duhaime (approved):

PR #1167, opened 2025-08-27
update the default icon marker to use the new one from chaire-lib

kaligrafy:
## Summary by CodeRabbit

* Style
  * Updated the default map marker icon for map point inputs to a new design, ensuring consistent appearance when no custom icon is set.
  * Standardized the fallback icon path across the app for improved reliability.

* Refactor
  * Centralized configuration for the default marker to reduce duplication and ease future updates.


kaligrafy:
Forget this, we need to change this in the survey, because evolution does not use the icons from chaire-lib yet.

PR #1173, opened 2025-08-27
security: escape strings entered by the participant

### LINKED ISSUE #1169: widgets: it is possible to inject javascript in inputs that can be processed later
If surveys have labels that contain HTML and those labels feed content from previous answers (like a place name), then, if that place name contains injected javascript code, it can be executed later.

We should make sure to escape any HTML tags from participant strings before saving to the server.

Ideally, we should also document evolution to make sure that survey that inject a participant response in a label, especially if it contains html, escape them before doing so.

That means, in the generator to escape the person's nickname.

tahini:

fixes #1169

In the backend, all values coming from the client are sanitized using `dompurify` to remove any injected javascript code.

In the generator, we make sure the escape the nickname using `lodash/escape` (which simply replaces special html characters by their code, like `<` becomes `&lt;`). This is an additional safeguard to our escaping in the backend

## Summary by CodeRabbit

* **New Features**
  * Interview updates now sanitize all string fields (including nested ones) before processing and surface cleaned values where applicable.
  * Widget configs now require localized label/text fields; a new translatable-string function type is supported.
  * Generated labels now escape nicknames when used.

* **Tests**
  * Added/updated tests validating sanitization, server-overwrite behavior, array handling, and nickname escaping.

* **Chores**
  * Added dependencies to support input sanitization.


samuel-duhaime (commented):
Question: Does any of these shortcuts HTML from the Generator will still work?

image

greenscientist (approved):
Looks good to me. 

About the need to add the package, shouldn't the generate generate its package list ? Having to handle that by hand seems a bit broken

tahini:
> About the need to add the package, shouldn't the generate generate its package list ? Having to handle that by hand seems a bit broken

Maybe, but we're not there yet. `package.json` is manual

PR #1188, opened 2025-08-28
Fixes the UI tests that input postal codes.

GabrielBruno24:
Because the postal codes fields now automatically add a space in the middle, this broke UI tests as the value read was different than the value entered. The inputString tests now have an optional expectedValue parameter, which will be  the expected value instead of the inputed value when present.

## Summary by CodeRabbit

- New Features
  - Postal code input now auto-formats to the standard pattern (e.g., "H1S1V7" → "H1S 1V7") for clearer display.

- Bug Fixes
  - Validation now accepts common entry variations and displays the normalized postal code instead of rejecting entries.

- Tests
  - UI tests updated to assert formatted postal code outputs and cover diverse input cases to ensure consistent normalization and display.

### DIFF
--- a/packages/evolution-frontend/tests/ui-testing/testHelpers.ts
+++ b/packages/evolution-frontend/tests/ui-testing/testHelpers.ts
@@ -44,6 +44,7 @@ type Path = string;
 type Email = string;
 type PathAndValue = { path: Path; value: Value };
 type PathAndValueBoolOrStr = { path: Path; value: StringOrBoolean };
+type OptionalExpectedValue = { expectedValue?: Value };
 type HasTitleTest = (params: { title: Title } & CommonTestParameters) => void;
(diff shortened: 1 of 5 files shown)


kaligrafy (approved):

kaligrafy:
You may check coderabbit's suggestions

packages/evolution-frontend/tests/ui-testing/testHelpers.ts
```diff
@@ -518,10 +519,10 @@ const inputString: InputStringTest = async ({ context, path, value }): Promise<L
 };
 
 // Test input string widget
-export const inputStringTest: InputStringTest = ({ context, path, value }) => {
+export const inputStringTest: InputStringTest = ({ context, path, value, expectedValue }) => {
```
greenscientist, line 522: Not required, but while you are at it, could you add a function header/description like the other function under it ?

PR #1189, opened 2025-08-29
survey: fix race condition in `startUpdateInterview` redux action

### LINKED ISSUE #1187: redux: race condition when running 2 widget updates on some browsers
Bisect in demo_survey shows the faulty commit is 446f2f2e190db0fb00d04c4f40041715855a0057

There appears to be a race condition when 2 widgets are updated at the same time on those browsers (for example, edit an input string, then click on a radio button). The next action does not wait for the previous one to be run, so the second one to run will use a version of the interview prior to the previous state update.

This does not seem to depend on the browser after all, as by faking a slow connection, it is possible to reproduce on a browser that otherwise shows no problem

tahini:
fixes #1187

With slower connections, many concurrent update operations can be run in parallel and the result of the subsequent will override the previous results. Update operations need to be run sequentially.

We used to have the `redux-async-queue` package to do this, but it does not work with latest version of redux that we used. It is not supposed to be necessary, according to redux documentation, but they also strongly suggest using `@reduxjs/toolkit`, which we have not done yet.

In the meantime, we need to run update in sequence, so we enqueue them in a 1-concurrent p-queue. We can investigate later if our use of redux is wrong, too complex or if the a new package would solve the problem.

Add a test for concurrent update calls. This test failed with previous code, but passes with this commit.

## Summary by CodeRabbit

- Bug Fixes
  - Serializes concurrent interview updates to prevent race conditions, ensuring actions, updates, and navigation occur in the correct order.

- Tests
  - Added comprehensive concurrency tests and improved test state handling to validate sequencing and outcomes under simultaneous updates.

- Chores
  - Added a lightweight queuing runtime dependency to support serialized asynchronous operations and improve stability.

### DIFF
--- a/packages/evolution-frontend/package.json
+++ b/packages/evolution-frontend/package.json
@@ -55,6 +55,7 @@
     "lodash": "^4.17.21",
     "moment": "^2.30.1",
     "node-fetch": "^2.6.7",
+    "p-queue": "^6.6.2",
     "react": "^19.0.0",
     "react-datepicker": "^7.6.0",
     "react-dom": "^19.0.0",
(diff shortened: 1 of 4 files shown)


tahini:
> I guess we need to test this a lot!

Yes we do!

samuel-duhaime (approved):

PR #1192, opened 2025-08-29
formatters: use custom access code postal code formatters and keep cursor position

tahini:

* the cleave-zen block formatting was not adequate for our use case, it was too agressive, or not enough, on the delimiter, not allowing users to either write or erase it. And the format is simple enough to have our custom functions

* Add a `FormattedInput` component that runs the inputFilter function and resets the previous position (or increments it to the right location) after the change, as otherwise, if the value changes, react or the browser automatically goes at the end of the string.


packages/evolution-common/src/utils/__tests__/formatters.test.ts
samuel-duhaime, line 1: I'm unsure, why the test are not the same anymore. Don't we want exactly the same result, even without the package?

samuel-duhaime (approved):
It looks good, can you make some test with Cellphone and tablet ?

PR #1197, opened 2025-08-31
Replace electric scooter by electric kick scooter

kaligrafy:
The name was incorrect, because electric scooter is not what we meant here.

Electric kick-scooter is a "Trottinette électrique" in french, which is now correct.

## Summary by CodeRabbit

* **Style**
  * Updated mode name from “E‑Scooter” to “Electric Kick Scooter” in English; short label now “E‑Kick Scooter”.
  * French labels remain “Trottinette électrique” while aligning terminology internally for consistency.
  * Refreshed the electric kick scooter icon to match the updated naming.
* **Chores**
  * Standardized mode terminology across settings and surveys to ensure consistent display and selection.

### DIFF
--- a/packages/evolution-common/src/services/baseObjects/attributeTypes/SegmentAttributes.ts
+++ b/packages/evolution-common/src/services/baseObjects/attributeTypes/SegmentAttributes.ts
@@ -16,7 +16,7 @@ export const modeValues = [
     'walk',
     'bicycle', // complementary type: BicycleType
     'bicyclePassenger',
-    'scooterElectric',
+    'kickScooterElectric',
     'carDriver', // complementary type: CarType
     'carPassenger',
     'transitBus', // includes Bus Transit System (BTS) category C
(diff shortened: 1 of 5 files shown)


packages/evolution-common/src/services/baseObjects/attributeTypes/SegmentAttributes.ts
```diff
@@ -56,7 +56,7 @@ export const mapModeToModeCategory: { [mode in Mode]: ModeCategory } = {
     walk: 'walk',
     bicycle: 'bicycle',
     bicyclePassenger: 'bicycle',
-    scooterElectric: 'bicycle',
+    kickScooterElectric: 'bicycle',
```
samuel-duhaime, line 59: C'est dans la catégorie bicycle?

kaligrafy: oui

PR #1220, opened 2025-09-15
admin: Update icon paths in the `InterviewMap` component

### LINKED ISSUE #1219: admin: missing icons in interviewMap
Now that we use icons from chaire-lib, the InterviewMap widget in the admin should also use those icons

tahini:
fixes #1219

The component should use the new icons from chaire-lib, with the icon helper function to retrieve the right value.


kaligrafy (approved):

kaligrafy:
Le path des icônes ne va pas changer mais tout le reste oui.

Issue #1268, opened 2025-10-07
Discussion: Should the validator be able to see the interview data when an interview is frozen.


samuel-duhaime:
# Issue

## Dicussion
@kaligrafy @tahini @greenscientist  Should the validator be able to see the interview data when an interview is frozen?

Right now, the validator cannot see the interview. It can only see that the interview is frozen because it goes to an 'unavailable' page. But maybe we want to see all the input in view mode only?

tahini:
More than that, there should be a discussion about what it means to be frozen. Currently, it was just implemented so that when an interview is explicitly set as valid or completed, then it is frozen. But is that the desirable behavior? Also, when should we set it? Too eager validators could cause an interview to be inaccessible.

* When should we make an interview not updateable anymore?
* Is is_frozen the right parameter for it?
* Is an interview frozen for the participant and interviewer only, or also for the validator?

Related to #1253

kaligrafy:
Admins and reviewers should always be able to change the validation status and review an interview multiple times.

kaligrafy:
As a first step, please disable frozen check in admin routes, otherwise it makes the review process unusable.

kaligrafy:
Replaced by #1949 #1950 #1967

PR #1271, opened 2025-10-08
Disable frozen check in admin/review mode

kaligrafy:
It made it impossible to load or edit an interview in review mode.

## Summary by CodeRabbit

- Bug Fixes
  - Users can now start the interview correction flow even when an interview was previously marked as frozen; the correction copy is created and audits run as usual.
  - Eliminates unexpected “Forbidden” errors during correction, ensuring a consistent, uninterrupted workflow and successful retrieval of interview data.
  - Improves reliability of correction requests and reduces friction for users performing post-interview updates.

### DIFF
--- a/packages/evolution-backend/src/api/survey.validation.routes.ts
+++ b/packages/evolution-backend/src/api/survey.validation.routes.ts
@@ -37,14 +37,6 @@ router.get(
             try {
                 const interview = await Interviews.getInterviewByUuid(req.params.interviewUuid);
                 if (interview) {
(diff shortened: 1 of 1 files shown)


samuel-duhaime (changes requested):
@kaligrafy @tahini @greenscientist  This is not a bug, it's a feature. 

We have a discussion in the issue https://github.com/chairemobilite/evolution/issues/1268 that you can contribute to before we make this decision. I'm still unsure of the best way to handle that. 

By the way, it was already impossible to load or edit an interview in review mode, it was showing an infinite loading state.

PR #1293, opened 2025-10-17
Add test to find orphan audit check locales

kaligrafy:
If any locale is set for an audit check that is no longer there, the test will fail. Makes it possible to keep locales in sync with audit check codes.

## Summary by CodeRabbit

* **Tests**
  * Added runtime validation to detect and fail on orphaned audit translation keys across locales, with logging for any discrepancies to ensure translations remain consistent with audit definitions.

### DIFF
--- a/packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/AuditLocales.test.ts
+++ b/packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/AuditLocales.test.ts
@@ -15,6 +15,7 @@ import { tripAuditChecks } from '../TripAuditChecks';
 import { segmentAuditChecks } from '../SegmentAuditChecks';
 import i18n, { registerTranslationDir, addTranslationNamespace } from 'chaire-lib-backend/lib/config/i18next';
 import path from 'path';
+import fs from 'fs';
 
 const languages = ['en', 'fr'];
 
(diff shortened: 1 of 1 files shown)


packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/AuditLocales.test.ts
```diff
@@ -70,4 +71,45 @@ describe('Audit Locales', () => {
...
+            describe.each(Object.entries(auditCheckModules))('Prefix: %s', (prefix, auditChecks) => {
+                it(`should not have orphaned translations starting with ${prefix}`, () => {
+                    const translationKeys = getTranslationKeysFromFile(language);
+                    const keysWithPrefix = translationKeys.filter((key) => key.startsWith(prefix));
+                    const auditCheckCodes = Object.keys(auditChecks);
+
+                    // Find translations that exist but are not in the audit check module
+                    const orphanedKeys = keysWithPrefix.filter((key) => !auditCheckCodes.includes(key));
+
+                    if (orphanedKeys.length > 0) {
```
tahini: Tu n'as pas besoin de ce bloc, le expect plus bas devrait te donner toute l'information dont tu as besoin: la chaîne orphanedKeys, le préfixe et la langue seront dans le nom du test.

tahini (approved):
Un petit console.log inutile à enlever, mais c'est bon

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/AuditLocales.test.ts
kaligrafy: removed!

PR #1334, opened 2025-12-10
Update transition submodule

GabrielBruno24:
Update transition submodule so we can use the new accessibility map API in localisation

## Summary by CodeRabbit

* **Chores**
  * Bumped minimum Node.js runtime to 24.x across the project and examples.
  * Updated dev type definitions (@types/node) to align with Node 24.
  * Switched CI workflows and test matrices to Node.js 24.x.
  * Updated Docker base image to Node 24.
  * Advanced an embedded submodule to a newer commit.

### DIFF
--- a/.github/workflows/evolution-generator.yml
+++ b/.github/workflows/evolution-generator.yml
@@ -13,7 +13,7 @@ jobs:
     runs-on: ubuntu-24.04 # As of Nov 2024, ubuntu-latest is too old for peotry
     strategy:
       matrix:
-        node-version: [22.x]
+        node-version: [24.x]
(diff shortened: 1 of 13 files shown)


GabrielBruno24:
Question: I ran `yarn` after updating the submodule to the latest commit. Is it normal that the yarn.lock file change, let alone to that extant? Does it update to match the dependencies in transition?

tahini:
It is normal that the yarn.lock changes a lot. Especially now we are updating packages in Transition, it may not be a good time to upgrade it in Evolution right now. Since it is not required , we should wait to do this later

PR #1353, opened 2026-01-07
paradata: log invalid widgets on button click

### LINKED ISSUE #1351: paradata: log invalid widgets on `buttonClick` events
The `buttonClick` event already saves the hiddenWidgets on the page, but we should also save the visible and invalid widgets.

Though issue #1350 adds a call to the server when trying to submit an invalid section, the `valuesByPath` may not necessarily show the valid/invalid widgets if the widget was already set as invalid previously. But when clicking a button and triggering an action, we do want to know which widgets were invalid at the moment of the action. We should thus add this information to the user event, like for the hidden widgets.

tahini:
fixes #1351

To properly track the fields that are invalid when the user clicks a
Button widgets to submit a page, they are added to the `buttonClick`
event. Otherwise, they may go unnoticed if it was previously set to
invalid, but never fixed. The validation status will not change and will
thus not be part of the `valuesByPath` data.

If there are `invalidWidgets` declared in a button_click userAction,
they are exported in addition to other invalid fields in the event's
valuesByPath

## Summary by CodeRabbit

* **New Features**
  * Admin export logs now include tracking of invalid widgets when users click buttons in surveys.

* **Tests**
  * Refactored button click event tests with parameterized test cases covering multiple widget scenarios.

### DIFF
--- a/packages/evolution-backend/src/services/adminExport/__tests__/exportInterviewLogs.test.ts
+++ b/packages/evolution-backend/src/services/adminExport/__tests__/exportInterviewLogs.test.ts
@@ -499,80 +499,107 @@ describe('exportInterviewLogTask', () => {
 
     });
 
-    test('Test with an event of type button_click with user action', async () => {
-        // Add one log statement, with/without hidden paths to test the button_click event:
-        const userAction = { type: 'buttonClick', buttonId: 'response.someField' };
-        const userActionWithHidden = { type: 'buttonClick', buttonId: 'response.otherField', hiddenWidgets: [ 'hiddenWidget1', 'hiddenWidget2' ] };
-        const buttonLogs: { [key: string]: any }[] = [{
-            ...commonInterviewData,
-            event_type: 'button_click',
-            timestamp_sec: 1,
-            event_date: new Date(1 * 1000),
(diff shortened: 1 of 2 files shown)


packages/evolution-frontend/src/actions/__tests__/Survey.test.ts
```diff
@@ -328,23 +328,25 @@ describe('Update interview', () => {
...
+                groupIdWithSomeInvisibleOrInvalid: {
+                    visibleWidget: { path: 'group.groupIdWithSomeInvisible.visibleWidget', isVisible: true, isValid: true },
+                    visibleInvalidWidget: { path: 'group.groupIdWithSomeInvisible.visibleInvalidWidget', isVisible: true, isValid: false },
+                    invisibleWidget: { path: 'group.groupIdWithSomeInvisible.invisibleWidget', isVisible: false, isValid: true }
                 }
             }
         } as any;
         const validatedInterview = _cloneDeep(expectedInterviewToPrepare);
-        validatedInterview.widgets = nextWidgetStatuses;
+        validatedInterview.widgets = nextWidgetStatuses as any;
```
samuel-duhaime: Aucun type pour cela ?

PR #1355, opened 2026-01-08
Bug: Fix bug with incorrect node version with tasks.json

samuel-duhaime:
# Pull request

## Description
Bug: Fix bug with incorrect node version with tasks.json

## Justification
Visual Studio Code terminal doesn't take the good Node version when doing tasks, but we need Node 24 for installing packages.

## Summary by CodeRabbit

* **Chores**
  * Updated development tooling configuration to ensure proper Node.js version compatibility during build processes.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -5,6 +5,7 @@
             "label": "Update Main",
             "dependsOrder": "sequence",
             "dependsOn": [
+                "NVM Use 24",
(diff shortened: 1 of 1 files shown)


tahini (commented):
I don't use this task file, so I'm not sure how it works, but coderabbitai's concern seems legit. You might prefer to set your default node version to 24 instead of requiring `nvm use` in each shell.

PR #1360, opened 2026-01-14
fix(ui): move zoom controls to top-left for better visibility

euzghe:
This PR moves the Google Maps zoom controls from RIGHT_BOTTOM to LEFT_TOP. This was requested by the maintainers in the localisation repository to ensure the controls are visible in smaller map containers. 
Reference: chairemobilite/localisation#54

## Summary by CodeRabbit

* **Style**
  * Relocated the map zoom control to the top-left, improving visibility and accessibility and reducing overlap with other UI elements.
  * This is strictly a UI positioning change; map behavior, interactions, and data flow remain unchanged.

### DIFF
--- a/packages/evolution-frontend/src/components/inputs/maps/google/InputMapGoogle.tsx
+++ b/packages/evolution-frontend/src/components/inputs/maps/google/InputMapGoogle.tsx
@@ -221,8 +221,7 @@ const InputMapGoogle: React.FunctionComponent<InputGoogleMapPointProps & WithTra
                 cameraControl: false,
                 zoomControlOptions: {
                     position: google.maps.ControlPosition.LEFT_TOP,
-                    style: google.maps.ZoomControlStyle.SMALL
-                }
+                     }
             }}
             onLoad={onLoad}
             onUnmount={onUnmount}


tahini:
The original issue is not as much about the placement as the fact that we need to click on the square control to see the zoom and other icons. The zoom buttons (+ / -) at least should be visible by default on the map, while panning buttons could be visible only on request. @samuel-duhaime can you confirm what is the expected behavior for the map controls?

euzghe:
Hi @tahini and @samuel-duhaime! I've updated the PR with the requested changes. I used google.maps.ZoomControlStyle.SMALL to ensure the '+' and '-' buttons are always visible as separate controls, even in smaller map containers. This prevents them from being collapsed into a single menu button. Thanks for the guidance!

euzghe:
Hi @tahini and @samuel-duhaime! I've removed the ZoomControlStyle line to fix the compilation errors as requested. However, as @samuel-duhaime mentioned, the main goal is to keep the zoom buttons (+/-) visible by default.

Since the library doesn't support the SMALL style, should we consider increasing the minimum height of the map container, or is there another preferred way in this project to prevent Google Maps from collapsing the controls into a square menu? I'm ready to implement the best alternative! 😊

PR #1366, opened 2026-01-16
segmentSection: decouple questionnaire mode type from model

### LINKED ISSUE #1363: segment section: decouple questionnaire type from base objects for mode
Currently, the mode and modePre widgets use the types from the `SegmentAttributes` in base object. But they should be decoupled as both server different purposes and are not fully compatible

* baseObjects related to the model that is built from the data collected in Evolution, so modes may be simpler and have additional fields to describe them. For example a `bicycle` mode could have a `bicycleType` field describing whether it is `electric` or `standard`, and another for whether it is privately owned or shared.
* questionnaire on the other hand should focus on reducing the respondent buden and not add additional questions. So all these questions could be answered by selecting a separate mode for the single mode questions. 

The enhancement process will then translate from one answer to the other.

With #1345, it should be possible de configure for each survey what level of data needs to be collected exactly.

tahini:
fixes #1363

The modes in `baseObjects` describe the model built by Evolution and thus a 'mode' may be divided in multiple fields, with the type of motorization, ownership, etc. In the questionnaire, though, we should limit the number of questions to ask the respondent to reduce the burden, so we ask one or two questions, mode category and optionally mode, with values that could be converted to multiple other fields.

This commit adds a `Mode` type on the questionnaire side and updates the builtin widgets to use those types. A few modes are added to the list.

Note that this is not exhaustive. Since the demo_survey uses those widgets and they are not yet configurable, every current mode is displayed. But, later commits will add segment section configuration, where it will be possible to specify the level of granularity in choices, with a few defaults that would limit and sort the choices available in different situations (for example transit-intensive urban areas vs rural areas). It will then be possible to add possible values from which to pick for a given survey.

## Summary by CodeRabbit

* **New Features**
  * Expanded transport modes (Paratransit, Ferry, CarDriver variants, PrivateBoat, Snowmobile, AllTerrainVehicle) with updated mode handling and icon support
  * New conditional checks for disability-related options to tailor choices

* **Localization**
  * Added English and French labels/translations for all new transport modes

* **Tests**
  * Added and updated unit tests covering disability conditionals and mode/label behavior

### DIFF
--- a/packages/evolution-common/src/services/questionnaire/sections/segments/__tests__/helpers.test.ts
+++ b/packages/evolution-common/src/services/questionnaire/sections/segments/__tests__/helpers.test.ts
@@ -377,3 +377,151 @@ describe('shouldShowSameAsReverseTripQuestion', () => {
     });
 
 });
+
+describe('conditionalPersonMayHaveDisability', () => {
+
+    test('Person has disability set to "yes"', () => {
+        // Prepare test data
+        const interview = _cloneDeep(interviewAttributesForTestCases);
+        interview.response._activePersonId = 'personId1';
+        setResponse(interview, 'household.persons.personId1.hasDisability', 'yes');
+
+        // Test conditional
+        const result = helpers.conditionalPersonMayHaveDisability(interview, 'household.persons.personId1');
+        expect(result).toEqual(true);
+    });
+
+    test('Person has disability set to "preferNotToAnswer"', () => {
+        // Prepare test data
+        const interview = _cloneDeep(interviewAttributesForTestCases);
+        interview.response._activePersonId = 'personId1';
+        setResponse(interview, 'household.persons.personId1.hasDisability', 'preferNotToAnswer');
+
+        // Test conditional
+        const result = helpers.conditionalPersonMayHaveDisability(interview, 'household.persons.personId1');
+        expect(result).toEqual(true);
+    });
+
+    test('Person has disability set to "dontKnow"', () => {
+        // Prepare test data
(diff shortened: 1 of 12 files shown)


packages/evolution-common/src/services/questionnaire/sections/segments/__tests__/widgetSegmentModePre.test.ts
```diff
@@ -159,6 +180,32 @@ describe('Mode choices conditionals', () => {
         expect(carDriverResult).toEqual(true);
     });
 
+    test('paratransit conditional should return true if household may have a person with disability', () => {
```
kaligrafy: Même chose ici, pas de lien avec Mode

kaligrafy (changes requested):

packages/evolution-common/src/services/questionnaire/sections/segments/__tests__/helpers.test.ts
```diff
@@ -377,3 +377,151 @@ describe('shouldShowSameAsReverseTripQuestion', () => {
     });
 
 });
+
```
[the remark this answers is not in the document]
tahini: ça a été déplacé dans ce fichier, car les fonctions sont utilisés dans 2 fichiers

packages/evolution-common/src/services/questionnaire/sections/segments/__tests__/widgetSegmentModePre.test.ts
kaligrafy: ok

PR #1379, opened 2026-01-30
accessibilityMap: add access/egress and walking speed parameters

tahini:
Those are parameters supported by Transition and that may be required by consumers of this function. Also type the properties of the returned polygons, to include the duration, area and extra accessibility data.

## Summary by CodeRabbit

* **New Features**
  * Added optional parameters to customize accessibility map calculations: max access/egress travel time and walking speed.
  * Accessibility map polygons now include richer metadata (duration, area, and optional counts of accessible places).

* **Bug Fixes**
  * Improved validation for parameter values and clearer error messages for bad requests.

* **Tests**
  * Expanded parameterized tests covering multiple scenarios (with/without POIs and optional params).

### DIFF
--- a/packages/evolution-backend/src/services/routing/types.ts
+++ b/packages/evolution-backend/src/services/routing/types.ts
@@ -89,6 +89,21 @@ export type AccessibilityMapCalculationParameter = {
     maxTotalTravelTimeMinutes: number;
     /** whether to calculate points of interest */
     calculatePois?: boolean;
+    /** The maximum access/egress travel time in minutes */
+    maxAccessEgressTravelTimeMinutes?: number;
+    /** The walking speed in kilometers per hour, to calculate access and egress travel times */
+    walkingSpeedKmPerHour?: number;
+};
+
+/**
+ * Properties for polygons in the accessibility map. Comes from the Transition
+ * API documentation
+ */
+export type AccessibilityMapPolygonProperties = {
+    durationSeconds: number;
+    areaSqM: number;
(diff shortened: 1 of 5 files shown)


packages/evolution-backend/src/services/routing/types.ts
```diff
@@ -89,6 +89,21 @@ export type AccessibilityMapCalculationParameter = {
     maxTotalTravelTimeMinutes: number;
     /** whether to calculate points of interest */
     calculatePois?: boolean;
+    /** The maximum access/egress travel time in minutes */
+    maxAccessEgressTravelTimeMinutes?: number;
+    /** The walking speed in kilometers per hour */
+    walkingSpeedKmPerHour?: number;
```
samuel-duhaime: Will we have a bikingSpeedKmPerHour ?
Or, this is a shortcut for now? If it's, we should add a Todo or a comment to said why we only have walkingSpeedKmPerHour.

samuel-duhaime:
@tahini Also, will we be able to have the api get the accessibility of 1 selected place or two places in the map?

Like in this screenshot?
image

packages/evolution-backend/src/services/routing/types.ts
tahini: No it's walking, and acutally, biking and driving are too fast, so turns out we can't really use this parameter the way I wanted to (we'll have to increase access/egress time instead in localisation)

kaligrafy (approved):

PR #1381, opened 2026-02-02
segment section: have a single object create section and widget config

tahini:
part of #501

Define 2 interfaces: `SectionConfigFactory` and `WidgetConfigFactory`, the former extending the latter. Config objects that provide more than one widget can use classes implementing those factories to return all the required widgets.

The `sectionSegments`, as well as `groupPersonTrips` and `groupSegments` use this pattern to advertise all their widgets and configs.

The demo_survey platform now gets the sections and widgets from a single `questionnaire.ts` file, which, for the segment section, now use the `SegmentsSectionFactory` to get all the builtin section config and widgets. Builtin widgets do not have to be added to the `segments.tsx` file anymore.

## Summary by CodeRabbit

* **Removals**
  * Removed person-switching UI and several segment-related widgets and exports.
* **Refactor**
  * Migrated section and widget definitions to a factory-based architecture; consolidated survey sections and widgets into a single questionnaire surface.
* **Changes**
  * Added an explicit segment section type and tightened function signatures requiring explicit section config.
* **Tests**
  * Updated and expanded tests to exercise the new factory APIs and added helpers for masking functions.

### DIFF
--- a/example/demo_survey/src/admin/app-admin.tsx
+++ b/example/demo_survey/src/admin/app-admin.tsx
@@ -2,8 +2,7 @@ import runClientApp from 'evolution-frontend/lib/apps/admin';
 import { setApplicationConfiguration } from 'chaire-lib-frontend/lib/config/application.config';
 import appConfig, { EvolutionApplicationConfiguration } from 'evolution-frontend/lib/config/application.config';
 
-import surveySections from '../survey/sections';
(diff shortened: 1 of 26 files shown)


kaligrafy (approved):

samuel-duhaime (changes requested):
Okay, c'est un bon début. Il manque quelques types à certains endroits.

Issue #1345, opened 2026-01-05
segment section: Provide a pluggable trips and segments section configuration that can be directly imported and configured in individual surveys


tahini:
This should include the mandatory widgets used by the trips and segments template.

By default, the section is absent, if it is defined in the survey, there should be default values set to the various configuration options.

Steps:

- [x] Provide a configurable mode list, where modes can be specified (included/excluded) for a survey
- [x] Do not require to individually list all widgets, groups and section, as currently, but let the section and widgets/groups all be configured by one single call with section configuration data.
- [x] Support adding additional questions in the segment section, that are not builtin.

PR #1386, opened 2026-02-06
segments: support adding custom widgets in segment group

tahini:
completes #1345

This adds the `additionalSegmentWidgetNames` configuration option to the `SegmentSectionConfiguration` type. It contains a list of widget names to add after the mode question and before the has next mode question.

These widgets need to be defined in the survey itself.

## Summary by CodeRabbit

* **New Features**
  * Survey segments can now include additional custom widgets via a new configurable list, allowing more flexible segment layouts.

* **Tests**
  * Added coverage verifying additional widgets are correctly included and ordered in segment configurations.

* **Chores**
  * Removed unused imports and cleaned up legacy commented code for clarity.

### DIFF
--- a/packages/evolution-common/src/services/questionnaire/sections/segments/groupSegments.ts
+++ b/packages/evolution-common/src/services/questionnaire/sections/segments/groupSegments.ts
@@ -24,6 +24,7 @@ export class SegmentsGroupConfigFactory implements WidgetConfigFactory {
     }
 
     private getSegmentsGroupConfig = (): GroupConfig => {
+        const additionalWidgetNames = this.sectionConfig.additionalSegmentWidgetNames || [];
         // TODO These should be some configuration receive here to fine-tune the section's content
         return {
             type: 'group',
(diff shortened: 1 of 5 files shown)


packages/evolution-common/src/services/questionnaire/sections/segments/groupSegments.ts
```diff
@@ -57,8 +58,9 @@ export class SegmentsGroupConfigFactory implements WidgetConfigFactory {
                 'segmentSameModeAsReverseTrip',
                 'segmentModePre',
                 'segmentMode',
+                // Additional widgets are added here
+                ...additionalWidgetNames,
```
samuel-duhaime: Will the additionalWidgetNames will always be exactly there or we have more control over the placement?

tahini:
The reason is that for the segment group, there is no real need for it. It would complicate how we configure it, I'd rather deal with that problem when it's required. Now, for the group segment, it goes there. For other sections, configurations may be different.

PR #1445, opened 2026-03-20
Add formating and linting on test files

tahini:
This adds overrides rules for the prettier and eslint rules for files in `__tests__` folders.

Using a line width of 150, a little longer than our usual 120, but ok for tests, I think.
Also, it tries to fit objects in one line with the `objectWrap: 'collapse'` (but honestly I have no idea if I prefer with `collapse` or `preserve`, you can try locally and let me know)

@kaligrafy @samuel-duhaime Can you both take a quick look at the test file formatting and let me know if something outrages you and we should try some more rules or if to have a proper formatting once and for all in test files is worth a bit of water in our wine.

When we agree on the rules, I'll put the rules in the official config files and document in the commit message the rules we agreed upon.

Also note that I upgrade version of `prettier` so there are a very few changes in files that are not in test folders (2 I think)

### DIFF
--- a/packages/evolution-backend/.eslintignore
+++ b/packages/evolution-backend/.eslintignore
@@ -1,3 +1,2 @@
 lib/
 node_modules/
-**/__tests__
--- a/packages/evolution-backend/.eslintrc.json
+++ b/packages/evolution-backend/.eslintrc.json
@@ -1,3 +1,13 @@
 {
-  "extends": "../../configs/base.eslintrc.json"
+  "extends": "../../configs/base.eslintrc.json",
+  "overrides": [
+        {
+            "files": ["**/__tests__/**/*.{ts,tsx}"],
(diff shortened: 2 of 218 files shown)


samuel-duhaime (commented):
Damn, ça fait un gros changements!
image

Pour ton premier avis, j'ai été voir les fichiers de test (ça me semblait mieux maintenant). Pour les règles, ça me semble aussi raisonnable. 
Ma philosophie de formatting, c'est de mettre le moins de règles possible (mais parfois, on n'a pas le choix).

samuel-duhaime:
T'en es où avec ce PR? :) @tahini

tahini:
> T'en es où avec ce PR? :) @tahini

Nulle part, c'était là pour discussion, on n'en a pas vraiment discuté! :D

PR #1470, opened 2026-04-01
visitedPlaces: add helper functions on visited places

tahini:
* Add a helper function to know if a visited place has a loop activity
* Add the `nextPlaceCategory` field in the VisitedPlace type
* Add the `getFirstIncomplteVisitedPlace` to get the first incomplete visited place, so it can be automatically selected
* Add the `reconcileVisitedPlaces` function to reconcile the visited place array after deletion/addition of the array (previously `updateVisitedPlaces` in the OD survey's code, like od_nationale_quebec)
* Add the `deleteVisitedPlace` function, which deletes a place, as well as the next home place if both previous and next were home. Also updates shortcut and reconciles the array after deletion.

These functions are required for the visited place template. They are pretty much copied and adapted as is from od_nationale_quebec. There's a lot of FIXME to explain the condition, possible limitations and potential improvements. As a first step for the visited place section, I would like to bring it as close to od_nationale_quebec as possible, to have the complete functionality in Evolution. When we do, we can discuss how to better update and re-use all those pieces, that should be better documented.

Knowing this, @samuel-duhaime @kaligrafy let me know if you see anything that should be addressed right now, otherwise, I'll continue the migration of the visited places template (there's another helper to add a new place that I'll need to port as well, then it's the tempalte tiself)

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for visited-place validation, including incompleteness scenarios, reconciliation logic, and deletion workflows.

* **Refactor**
  * Consolidated loop-activity detection by introducing a dedicated helper function, replacing direct membership checks across the codebase.
  * Enhanced visited-place data structure with a new `nextPlaceCategory` field for improved trip routing and categorization.

### DIFF
--- a/packages/evolution-common/src/services/odSurvey/__tests__/helpers.test.ts
+++ b/packages/evolution-common/src/services/odSurvey/__tests__/helpers.test.ts
@@ -13,6 +13,7 @@ import projectConfig from '../../../config/project.config';
 import { Journey, Person, Trip, UserInterviewAttributes, VisitedPlace } from '../../questionnaire/types';
 import { setProjectConfiguration } from 'chaire-lib-common/lib/config/shared/project.config';
 import { loopActivities } from '../types';
+import { otherPlace2P1Coordinates } from '../../../tests/surveys/testCasesInterview';
 
 const baseInterviewAttributes: Pick<
     UserInterviewAttributes,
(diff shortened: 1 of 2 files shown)


samuel-duhaime (approved):
Ça me semble correct, mais @kaligrafy va avoir une opinion plus technique que moi sur le sujet.

PR #1524, opened 2026-04-23
doc: replace `jsdoc` with `typedoc`

### LINKED ISSUE #1523: doc: use `typedoc` instead of `jsdoc` for documentation
`jsdoc` uses the javascript files to generate the documentation, effectively loosing all types from the generation. We want to use the typescript code directly instead. The `typedoc` package seems to do what we want, in addition to making prettier site, so we should switch to that https://www.npmjs.com/package/typedoc

tahini:
fixes #1523

`typedoc` uses the typescript source files to generate the
documentation, so it includes all the types, that were removed from the
`jsdoc` documentation. It incidentally also produces prettier
documentation by default.

As previously, the doc only includes the evolution-common package,
excluding test files. It sorts the content of files alphabetically and
creates the doc at the same location as before.

The doc generation script is now `yarn docs:api` to differentiate from
the previous which required to install manually a package. The `typedoc`
package is added as a dev dependency, as it may be generated by the CI
for publication

The default entry page of the doc is the repo's main README, which we
may want to override to customize the main text.

Also update a few @returns tag, that were undefined and threw errors at generation time

## Summary by CodeRabbit

* **Documentation**
  * Migrated internal API documentation generation process
  * Improved documentation comments across service modules for clarity

* **Chores**
  * Updated documentation generation configuration and scripts


samuel-duhaime:
@tahini I was trying to test, but I got this error in src/services/questionnaire/sections/visitedPlaces/widgetsGeography.ts:58:97

```
error TS2345: Argument of type '["visitedPlaces:LocationNameAddressExample" & { readonly [$SelectorKeyBrand]: unique symbol; }, { context: unknown; }]' is not assignable to parameter of type '[key: ("visitedPlaces:LocationNameAddressExample" & { readonly [$SelectorKeyBrand]: unique symbol; }) | ("visitedPlaces:LocationNameAddressExample" & { ...; })[], options?: (Omit<...> & { ...; }) | undefined] | [key: ...] | [key: ...]'.
  Type '["visitedPlaces:LocationNameAddressExample" & { readonly [$SelectorKeyBrand]: unique symbol; }, { context: unknown; }]' is not assignable to type '[key: string | string[], defaultValue: string, options?: ({ readonly context: unknown; } & $Dictionary) | undefined]'.
    Type at position 1 in source is not compatible with type at position 1 in target.
      Type '{ context: unknown; }' is not assignable to type 'string'.

58                     ? `<span class="_pale _oblique">(${t('survey:forExampleAbbreviation')}: ${t(key, { context: activity })})</span>`
```

samuel-duhaime (approved):
Never mind, this is working since your last change. I tested the HTML code, this look much better than the JSDoc HTML. :) 

Just a comment. So does it read the type or the comment on top of the function? It looks like it read both pieces of information. Can you config that?

tahini:
> Just a comment. So does it read the type or the comment on top of the function? It looks like it read both pieces of information. Can you config that?

What do you mean? Do you have an example? It reads the jsdoc on top of everthing I think, so I'm not sure I understand

samuel-duhaime:
@tahini 

No, not exactly, it's a bit different than before. In this example, there is not even a TypeDoc on top of the function.

image

image

Issue #1347, opened 2026-01-05
segment section: Document all possible configuration points.


tahini:

samuel-duhaime:
@tahini What do you mean exactly?

tahini:
A README somewhere should document the expected segment section configuration when it is defined, so that a user can just read the readme and understand how to properly configure the segment section.

PR #1525, opened 2026-04-23
doc: add an entrypoint for the internal API documentation

tahini:
Add a `docs/surveyAdminDoc.md` file, that will be the entrypoint readme of the Evolution API documentation. This page is meant to be read in the site generated by the `typedoc` package and targets the survey designers, with instructions on how to implement a new survey. It is not a developer documentation, nor a user documentation.

Move the README.md from `packages/evolution-generator` to the docs, under the `generator.md` name, so it can be added and linked to by the typedoc documentation.


tahini:
Draft for now, questions to consider:

1. @samuel-duhaime if we move the generator README "here", in the docs, is it ok?
2. Is it ok to put this in the `docs` folder or should we differentiate the "Evolution" documentation (with concepts, survey methodology, nomenclature, info on audits, info aimed at developers) from the "survey designer" doc, whose audience are people wanting to create a new survey with Evolution? And add a different folder for this?
3. The markdown files modified here contain links that would be broken if clicked on from the Evolution github repo, but would work only in the context of the typedoc documentation. I don't know how we can reconcile that. TBD, typedoc may have a solution.
4. Shall we provide code example, or place holders in the markdown that can be filled automatically by code that is certain to compile (for example for the QuestionnaireFactory code, we could use the demo_survey files with some tags to highlight the parts to put here). We're lazy, we don't want to maintain doc. But that can come later, a first version could contain simply no code.

tahini:
Aussi @samuel-duhaime  est-ce que rouler le générateur dépend de la présence du fichier README.md pour vrai?

tahini:
This would fix #1347

samuel-duhaime (approved):
@tahini Yes, it's okay, just make sure to change the readme path from pyproject.toml. 

NB: you could probably also change my name from the author list. I'm not the only one now.

image

PR #1527, opened 2026-04-24
ci: add a backport to 0.5.0 workflow for merged PRs

tahini:
This workflow automatically creates a PR against the v0.5.0 branch after merging a PR that is set to be backported.

PRs to backport are the ones labeled with `dependencies` (security updates by dependabot) and `backport-v0.5.0`, a label added manually to the PR.

The github-actions bot will automatically try to create a branch, based on the ref v0.5.0 and cherry-pick all the commits from the merged PR. This works only for PRs merged with the "rebase and merge" strategy, that we mostly use.

If a conflict is encountered while cherry-picking, it will add a comment in the merged PR to tell to do the backporting manually, otherwise, it will create a PR with the new branch, using the same title, prefixed with `[Backport v0.5.0]`.

Commit developed with the help of Copilot

## Summary by CodeRabbit

* **Chores**
  * Added a reusable CI workflow that creates uniquely named branches, validates merge commits, attempts commit-range cherry-picks, captures conflicts (and posts manual-resolution instructions), pushes successful branches, and opens PRs while avoiding duplicates.
  * Added a triggering workflow that invokes the reusable workflow across configured maintenance branches.

### DIFF
--- a/.github/workflows/backport-generic.yml
+++ b/.github/workflows/backport-generic.yml
@@ -1,0 +1,171 @@
+# Reusable workflow to backport PRs to a specific target branch (e.g.
+# maintenance branch).
+#
+# Note that other repositories can re-use the generic backport workflow by
+# defining a workflow that uses
+# chairemobilite/evolution/.github/workflows/backport-generic.yml@main workflow.
+name: Backport To Specific Target Branch
+
+on:
+  workflow_call:
+    inputs:
+      target_branch:
+        required: true
+        type: string
+
+# CAUTION: Since this workflow is run on pull_request_target, it has access to
+# the full repository and can run any code in it with write permissions. It
+# should NEVER checkout code from the PR branch directly, as it could be
+# malicious. We should only cherry-pick merged code from the main branch. See
+# https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
+jobs:
+  backport:
+    # Run only when the PR was merged in main and labeled for backporting.
+    # Accepted labels: backport-<TARGET_BRANCH> or backport-all.
+    if: |
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.base.ref == 'main' &&
+      (
+        contains(github.event.pull_request.labels.*.name, format('backport-{0}', inputs.target_branch)) ||
+        contains(github.event.pull_request.labels.*.name, 'backport-all')
+      )
+    runs-on: ubuntu-latest
+    permissions:
+      # Needed to push the backport branch.
+      contents: write
+      # Needed to open the backport PR.
+      pull-requests: write
+    env:
+      TARGET_BRANCH: ${{ inputs.target_branch }}
+    steps:
+    - name: Checkout target branch
+      uses: actions/checkout@v4
+      with:
+        # Start from the maintenance branch we backport to. This is a trusted reference since only maintainers can merge PRs to main, and the workflow only runs on merged PRs.
+        ref: ${{ env.TARGET_BRANCH }}
+        # Full history is safer for cherry-pick operations.
+        fetch-depth: 0
+
+    - name: Cherry-pick PR commits to a backport branch
+      id: backport
+      env:
+        PR_NUMBER: ${{ github.event.pull_request.number }}
+        TARGET_BRANCH: ${{ env.TARGET_BRANCH }}
+        # This workflow assumes "Rebase and merge": merge_commit_sha is then the last rebased
+        # commit on the base branch, and the N previous commits (N = pull_request.commits) are
+        # exactly the PR's commits. We use MERGE_SHA~N..MERGE_SHA to cherry-pick them.
+        # NOTE: Does not work for squash merges (merge_commit_sha would be a single squashed
+        # commit, so MERGE_SHA~N..MERGE_SHA would reach into unrelated base-branch history).
+        MERGE_SHA: ${{ github.event.pull_request.merge_commit_sha }}
+        PR_COMMITS: ${{ github.event.pull_request.commits }}
+      run: |
+        set -euo pipefail
+
+        BRANCH="backport/pr-${PR_NUMBER}-to-${TARGET_BRANCH}"
+        echo "branch=${BRANCH}" >> "$GITHUB_OUTPUT"
+
+        # Standard bot identity recommended by GitHub Actions docs. 41898282 is the github-actions[bot] user id. (see https://github.com/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token)
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+        # Fetch the base branch (main) so MERGE_SHA is available locally.
+        git fetch origin main
+
+        git checkout -b "$BRANCH"
+
+        # Ensure the merge commit is a rebase merge (1 parent). Evolution never
+        # allows merge commits, but another repo using this workflow could and
+        # we don't support it.
+        PARENT_COUNT=$(git cat-file -p "$MERGE_SHA" | grep -c '^parent ')
+        if [ "$PARENT_COUNT" -ne 1 ]; then
+          echo "Expected rebase merge (1 parent), got $PARENT_COUNT parents. Squash/merge-commit not supported." >&2
+          exit 1
+        fi
+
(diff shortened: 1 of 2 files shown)


tahini:
@greenscientist je n'ai pas testé ce workflow (évidemment) et copilot est pas mal l'auteur de tout ça, quoiqu'on a fait plusieurs itérations avant d'arriver à cette version finale et j'ai tout double-checké. Ça ne m'a pas l'air bien bien dangereux, mais svp double check pour être sûr qu'on se retrouve pas à poster qqc dans un repo random en qqc part (comme qqn a l'air d'avoir fait récemment dans Transition...)

.github/workflows/backport-maintenance.yml
```diff
@@ -0,0 +1,148 @@
+name: Backport To Maintenance Branch
+
+on:
+  # pull_request_target runs in the base repository context and can write PRs/comments.
+  pull_request_target:
+    types: [closed]
```
greenscientist: Might not be pertinent,  but is there a way to limit that to merged PR, not just closed one? (Sometimes we close without merging)

.github/workflows/backport-maintenance.yml
```diff
@@ -0,0 +1,148 @@
...
+
+env:
+  # Name of the maintenance branch we want to backport to.
+  TARGET_BRANCH: v0.5.0
```
greenscientist: Eventually, you will want to put that in a matrix style thing, since we will have multiple branch to backport to.

.github/workflows/backport-maintenance.yml
greenscientist: Ok I see that it's filtered later, maybe mention it in the comment.

.github/workflows/backport-maintenance.yml
```diff
@@ -0,0 +1,148 @@
...
+
+jobs:
+  backport:
+    # Run only when the PR was merged in main and labeled for backporting.
+    # Accepted labels: backport-<TARGET_BRANCH> or dependencies.
+    if: |
+      github.event.pull_request.merged == true &&
+      github.event.pull_request.base.ref == 'main' &&
+      (
+        contains(github.event.pull_request.labels.*.name, format('backport-{0}', env.TARGET_BRANCH)) ||
```
greenscientist: Eventually, I think you'll want a backport "all" tag. With the matrix comment I said above.

.github/workflows/backport-maintenance.yml
```diff
@@ -0,0 +1,148 @@
...
+        contains(github.event.pull_request.labels.*.name, 'dependencies')
```
greenscientist: Feels like it should be backport-dependencies to make it more explicit.

.github/workflows/backport-maintenance.yml
```diff
@@ -0,0 +1,150 @@
...
+      )
```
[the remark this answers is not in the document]
greenscientist: @tahini Seems like the rabbit has a point. Maybe need to double check the documentation. Maybe my suggestion fix that.

.github/workflows/backport-generic.yml
```diff
@@ -0,0 +1,148 @@
...
+        PR_COMMITS: ${{ github.event.pull_request.commits }}
+      run: |
+        set -euo pipefail
+
+        BRANCH="backport/pr-${PR_NUMBER}-to-${TARGET_BRANCH}"
+        echo "branch=${BRANCH}" >> "$GITHUB_OUTPUT"
+
+        # Standard bot identity recommended by GitHub Actions docs. (see https://github.com/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token)
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
```
greenscientist, line 69: Where does 41898282 comes from

.github/workflows/backport-generic.yml
```diff
@@ -0,0 +1,148 @@
...
+
+          const existing = await github.rest.pulls.list({
+            owner,
+            repo,
+            state: 'open',
+            base,
+            head
+          });
+
+          if (existing.data.length > 0) {
```
greenscientist, line 149: Should we do this existing check before we attempt the cherry-pick 2 steps above??

.github/workflows/backport-generic.yml
```diff
@@ -0,0 +1,148 @@
...
+            `Backport of #${sourcePr.number} to ${base}.`,
+            '',
+            'This PR was created automatically by the backport workflow.'
+          ].join('\n');
+
+          await github.rest.pulls.create({
+            owner,
+            repo,
+            base,
+            head: process.env.BACKPORT_BRANCH,
```
greenscientist, line 168: Is it normal that we don't reuse the head above??  Maybe the other head variable should be named more precisely? (This might need a comment)

greenscientist (commented):
Several comments, the rabbit have some point, that you may have already fixed (I see that you did an update while I was reading this.)

greenscientist:
instead of doing a split file like you did, please investigate doing it matrix style.

.github/workflows/backport-maintenance.yml
tahini: See new version, the env is not available in the `if` of the job, so we use a re-usable workflow with inputs instead. The specific one is much simple, so yeah, we could eventually have a matrix, but we can start with a simple one for now, to test it

.github/workflows/backport-maintenance.yml
tahini: Or we can just add the proper label to the dependencies PR if we don't want it automatic. I wanted one less manual step, but ok, I'll drop the depencies if you wish.

.github/workflows/backport-generic.yml
tahini: > Where does 41898282 comes from

Great question! That I also had seeing this, and that's why I pointed to the doc in the comment above, it's the github-actions bot user id

.github/workflows/backport-generic.yml
tahini: 2 steps above in the job or in the script. coderabbit suggests to force-push the branch instead, effectively overriding whatever was there before, so existing PRs will be updated automatically with the new one. Also 2 steps above is cherry-pick, while here, this step involves the PR, so I think it's ok to do the check here for existing PR

tahini (commented):
Answers to some of your comments, see updated version for a re-usable workflow (that will even avoid duplication in our survey repos)

tahini:
> instead of doing a split file like you did, please investigate doing it matrix style.

It's not instead, it's in addition to (I'm assuming you talk about the newer version with re-usable workflow). The reusable workflow remains for 1- receiveing the branch as inputs so it's available at job-level if 2- being directly re-usable in our other repos, without duplicating the file there.

For now, we only have v0.5.0 (which should really be v0.5.x as there's a single branch). If you insist I can do it matrix style, or keep it as is for now.

.github/workflows/backport-generic.yml
```diff
@@ -0,0 +1,152 @@
...
+        # Standard bot identity recommended by GitHub Actions docs. 41898282 is the github-actions[bot] user id. (see https://github.com/actions/checkout#push-a-commit-to-a-pr-using-the-built-in-token)
+        git config user.name "github-actions[bot]"
+        git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+
+        # Fetch the base branch (main) so MERGE_SHA is available locally.
+        git fetch origin main
+
+        git checkout -b "$BRANCH"
+
+        if ! git cherry-pick -x "${MERGE_SHA}~${PR_COMMITS}..${MERGE_SHA}"; then
```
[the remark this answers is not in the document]
tahini: It's fine, we usually use rebase and merge, and it's documented, we never do merge commits

greenscientist:
Can you add this information directly in the generic file. 

"Note that other repositories can re-use the generic backport workflow by
defined a workflow that uses
`chairemobilite/evolution/.github/workflows/backport-generic.yml@main`
workflow."

It's weird to have a generic file with only one user in the repo, so adding an explaination of why it's separate in the top of the file will be useful, and people will be less tempted to copy the whole thing. (People will not see that commit message when creating their repo)

greenscientist (approved):
good comment addition. Feels less likely that our future people will mess up

tahini:
> good comment addition. Feels less likely that our future people will mess up

Hopefully our future selves will read comments in random places of a file (unlike my current or past self...) or at least some ai bot will! We will have been warned!

PR #1532, opened 2026-04-27
generator: generate labels with nested levels of indentation

### LINKED ISSUE #1529: generator: expand label keys to produce objects in yaml file
Evolution's locales file use an object-based approach, with child keys added under a parent key, like this:

```
mode:
    Walk: <strong>Marche</strong>
    Bicycle: <strong>Vélo</strong>
```

The generator produces dot-separated keys in the yaml file like this:

```
Mode.Walk: <strong>Walking</strong>
mode.Bicycle: <strong>Bicycle</strong>
```

The intention is that the generator overrides Evolution's labels. But the i18next webpack loader does not merge them magically, so that it is not effectively overridden. The generator should generate labels in yaml files in the same format as in Evolution, for the override to work.

The `choices` labels are correctly handled (though it works at only 1 level), all the other labels should be handled similarly. And the choices should use the same code path as the rest, as visibly it doesn't.

tahini:
fixes #1529

To match Evolution's locale files structure and ensure that labels generated by the generator effectively override Evolution's even if they include nested levels, we unnest the labels by splitting on '.' characters.

This is actually the behavior that was implemented for the choices yaml file, so it can now use the default function as well.

Add tests for the save_translations function of the `LabelsGenerator` class.

## Summary by CodeRabbit

* **Refactor**
  * Translation output now produces nested YAML keys (dot-paths expanded) and delegates choices locale persistence to the central labels generator.
  * Saves merge into existing files without duplicating headers and skips writing when a section has no translations.

* **Tests**
  * Added comprehensive tests for nested YAML generation, merge behavior, conflict handling, header preservation, and no-write-on-empty behavior.

* **Documentation**
  * Changelog updated to note nested-key YAML output.


samuel-duhaime (approved):
Okay, good!

tahini:
Added a changelog before merging.

PR #1544, opened 2026-04-28
questionnaire: extract a `yesNoChoices` array

tahini:
This is a simple 2 values array of choices for a yes/no question. The resulting value type is a boolean, `true` for `yes` and `false` for `no`. The current widgets using a yes/no choice are updated to use this choices array.


samuel-duhaime (changes requested):
Where is the translations files?

tahini (commented):
No translation file needed, it's already there, the choices are just extracted

samuel-duhaime (approved):

PR #1548, opened 2026-05-01
widgetConfig: better document types and rename a property

tahini:

- Document a few properties of the widgetConfig to clarify
- Add the `lastAction` property to geojson properties
- Rename `updateDefaultValueWhenResponded` to `resetToDefaultUnlessUserInteracted`, which is more on par with the behavior of the property. Extract the part checking it in the `WidgetOperation` and add unit tests for it.

## Release Notes

* **Breaking Changes**
  * Map widget property `updateDefaultValueWhenResponded` renamed to `resetToDefaultUnlessUserInteracted` for clearer intent and broader applicability across map types.
  * Map widget default value format updated to include enhanced metadata structure.

* **New Features**
  * Map widgets now track user interaction type to improve reset behavior determination.


kaligrafy (approved):
yes!

PR #1561, opened 2026-05-08
tripDiary: add the input times widgets

### LINKED ISSUE #1069: tripDiary: Make time widgets parameterizable and bring them to evolution
I just copy-pasted 5 different `InputTimeWidget` for a survey. They are quite similar, they mostly look at the current and previous places in their functions. The boundaries of the times are different, as well as the labels (mostly complex because of loop activities handling).

When moving the trip diaries to Evolution, this is the kind of widgets that we can make in one version, with configurable data.

tahini:
fixes #1069

add a `VisitedPlaceTimeWidgetFactory` for time widgets. It adds the 5 time widgets: previousPreviousDeparture, PreviousArrival, previousDeparture, arrival and departure times, with exhaustive unit test coverage of conditionals, labels, etc. Labels to override for each, as well as the possible specific labels to implement is documented in the factory's jsdoc.

The widgets are added to the questionnaire factory.

It is a very big PR, but it reduces the original code for those 5 widgets from ~1000 lines of custom code, to simpler, documenter, modular functions. We'll probably hit some bugs still, but at least it's a start and it's in Evolution, fully tested.

## Summary by CodeRabbit

* **New Features**
  * Expanded visited-places time inputs with multiple prior-arrival/departure options and richer arrival/departure flows
  * Added formatted trip-duration helper to produce human-friendly duration text in the UI

* **Localization**
  * Added extensive English and French translations for time prompts, labels, shortcuts, and trip-duration wording

* **Bug Fixes**
  * Reordered time fields for clearer, more consistent sequencing in the visited-places flow

* **Tests**
  * Updated and added tests to cover new time widgets and trip-duration formatting

[Review Change Stack](https://app.coderabbit.ai/change-stack/chairemobilite/evolution/pull/1561)


locales/en/visitedPlaces.yml
```diff
@@ -122,4 +122,84 @@ alreadyVisitedBySelfOrAnotherHouseholdMember: >-
...
+    At what time did {{nickname}} leave home to go to school?
+visitedPlacePreviousDepartureTime_home_schoolUsual_one: >-
+    At what time did you leave home to go to school?
+visitedPlacePreviousDepartureTime_home_other: >-
+    At what time did {{nickname}} leave home to go to this place ({{visitedPlaceDescription}})?
+visitedPlacePreviousDepartureTime_home_other_one: >-
+    At what time did you leave home to go to this place ({{visitedPlaceDescription}})?
+
+tripDurationText: <br /><span class=\"_pale _oblique\">({{durationText}} trip)</span>
+tripDuration:
```
samuel-duhaime, line 173: Maybe this is a good case to use i18n relativeTime?

https://www.i18next.com/translation-function/formatting#relativetime

samuel-duhaime (approved):
Un petit commentaire.

locales/en/visitedPlaces.yml
tahini: > Maybe this is a good case to use i18n relativeTime?
> 
> https://www.i18next.com/translation-function/formatting#relativetime

I don't think so, relative time is relative to some date, assignedDate would be such a use case (X days ago), but here it's a duration, (arrival - previous departure) so it does not apply

PR #1573, opened 2026-05-12
generator: use the widgetName instead of path as label key

### LINKED ISSUE #1530: generator: use widget name as translation key instead of path
The section/widgetName pair are expected to be unique in the generator. Plus the widgetName is usually a simple string, without dots or anything.

The path, on the other hand, may not exist (`TextWidgetConfig` does not need a path for instance), may take complex forms like `household.persons.{_activePersonId}.journeys.{_activeJourneyId}.departurePlaceOther`, which is not a very friendly translation key. Also, different groups on a same segment could technically have the same paths, in which case the key would clash (didn't happen yet in our surveys, but could happen).

We should thus switch to using widgetName for the translation key.

tahini:
fixes #1530

Rename the `path` parameter in the translation functions to `label_key` which represents more that the data is.

Use the `questionName` header field in priority to the `path` if it exists. Otherwise fallback to path for label key.

## Summary by CodeRabbit

* **Refactor**
  * Translation keys now derive from question names; spreadsheet label/namespace columns are configurable and gendered/base label variants are handled consistently.

* **Tests**
  * Expanded coverage for spreadsheet-to-translation generation and updated assertions to reflect new label key behavior.

* **Documentation**
  * Changelog updated with breaking-note guidance on renamed sheet columns and new label-key behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chairemobilite/evolution/pull/1573)


packages/evolution-generator/src/scripts/labels_generator.py
```diff
@@ -431,7 +433,11 @@ def add_translations_from_excel(
 
             # Find the index
             section_index = widgets_headers.index("section")
-            path_index = widgets_headers.index("path")
+            label_key_index = (
+                widgets_headers.index("questionName")
+                if "questionName" in widgets_headers
```
samuel-duhaime: questionName is required, no? It should be there, no?

packages/evolution-generator/src/tests/test_labels_generator.py
```diff
@@ -106,6 +108,168 @@ def test_get_labels_file_path_widgets():
...
 
 
+# Path where create_mocked_excel_data writes the workbook; we delete it after each test.
+MOCKED_EXCEL_FILE = "src/tests/references/test.xlsx"
+
+
+class TestAddTranslationsFromExcel:
+    """Tests for add_translations_from_excel function"""
+
+    @pytest.mark.parametrize(
```
samuel-duhaime, line 120: Can you use a case with parameters on each line? Hard to follow otherwise.

samuel-duhaime (commented):
Small modification.

Also, do we need to change some key in customLabel.yaml? I don't think so, but maybe it should be in the CHANGELOG for some survey.

samuel-duhaime (changes requested):

packages/evolution-generator/src/scripts/labels_generator.py
tahini: Not in the "Labels" sheet, the name is 'path', but maybe we should send as parameter the header to use as key

packages/evolution-generator/src/scripts/generate_survey.py
```diff
@@ -145,9 +145,16 @@ def generate_survey(config_path):
         # TODO: At some point, we should consider only read the Excel sheet one time, to avoid reading it multiple times.
         # TODO: We might consider extracting the sheet names from the Excel file or config file instead of hardcoding them.
         # Generate the labels for the specified sheets
-        sheet_names = ["Widgets", "Labels"]
+        sheet_names = [
```
samuel-duhaime: Is sheet_names really the good const name now?

samuel-duhaime (commented):
Just a 'const' name to update.

samuel-duhaime (approved):

PR #1576, opened 2026-05-12
Generator: Add section configuration generation with custom conditionals

### LINKED ISSUE #997: Generator: support custom conditionals for section configs function
The SectionConfig's `enableConditional` and `completionConditional` are now hard-coded in relation to the current or previous's section's completion status in the responses.

While this is good default for the navigation in the menu bar, with the upcoming navigation service that needs to not only verify a section's completion status, but also validate it, for a specific iteration of a section (for example the visited places of person X), most surveys currently have more complex completion/validation functions (for example `householdMemberIsComplete`, or `visitedPlacesForPersonIsComplete`).

The new navigation will also have additional properties to determine if a section should be skipped or is all its data valid. In an ideal world, we could all define that declaratively in config, but it is not the case yet, so in the meantime, it would be nice for the generator's section configuration to support customConditionals, similar to the widgets.

samuel-duhaime:
# PR

## Description
Fixes: #997
- Updated SectionConfig to support WidgetConditional for enableConditional and completionConditional properties.
- Add the new columns in the demo-generator Excel file.

## Summary by CodeRabbit

* **New Features**
  * Section configurations now support widget-aware conditional parsing for both section enablement and completion logic, extending the range of available conditional evaluation strategies.

* **Tests**
  * Comprehensive test coverage added for conditional expression mapping and section configuration generation, including parametrized scenarios covering various conditional configuration options.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chairemobilite/evolution/pull/1576)


packages/evolution-common/src/services/questionnaire/types/SectionConfig.ts
```diff
@@ -222,7 +222,7 @@ export type SectionConfig = {
      * FIXME: As we are moving towards a config-only section definition, this
      * function should be replaced with something that can be defined in config
      */
-    enableConditional?: boolean | ParsingFunction<boolean>;
+    enableConditional?: boolean | ParsingFunction<boolean> | WidgetConditional;
```
tahini: Ici, tu peux remplacer `ParsingFunction<boolnea>` par `WidgetConditional`. Le enableConditional est appelé à l,intérieur de la fonction `parseBoolean` qui prend ce type.

packages/evolution-common/src/services/questionnaire/types/SectionConfig.ts
```diff
@@ -232,7 +232,7 @@ export type SectionConfig = {
      * FIXME: As we are moving towards a config-only section definition, this
      * function should be replaced with something that can be defined in config
      */
-    completionConditional?: boolean | ParsingFunction<boolean>;
+    completionConditional?: boolean | ParsingFunction<boolean> | WidgetConditional;
```
tahini: idem

tahini (approved):

PR #1609, opened 2026-05-20
generator: add command line argument to override enabled scripts

tahini:
fixes #1601

When developing a questionnaire, we may want to build only part of it, for example the choices, labels or widgets for quick development. Long questionnaire can take a while to generate everything and it is not desirable to do it at all times.

This adds the command line argument `--only` that takes a comma-separated list of scripts to generate. For example, to generate only the widget configs, conditional and labels, adding `--only widgets_configs,widgets,conditionals,labels` would generate only those data, ignoring the rest.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line option to generate only selected scripts instead of running the full generation set.

* **Documentation**
  * Updated usage guidance with an example showing how to pass the selection via the new command-line option.

[Review Change Stack](https://app.coderabbit.ai/change-stack/chairemobilite/evolution/pull/1609?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_survey.py
+++ b/packages/evolution-generator/src/scripts/generate_survey.py
@@ -3,7 +3,9 @@
 # License text available at https://opensource.org/licenses/MIT
 
 # Note: This script generate the survey with multiple scripts.
-# These functions are intended to be invoked with the config YAML file.
+# These functions are intended to be invoked with the config YAML file.  They
+# can be overridden with the --only argument to only run a subset of the
+# scripts, which is useful for development and debugging.
 import argparse  # For command-line arguments
 from dotenv import load_dotenv  # For environment variables
 import os  # For file operations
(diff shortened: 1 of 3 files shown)


packages/evolution-generator/src/scripts/generate_survey.py
```diff
@@ -23,10 +23,74 @@
...
+    "generate_choices",
+    "generate_input_range",
+    "generate_labels",
+    "generate_UI_tests",
+    "generate_questionnaire_list",
+    "generate_questionnaire_dictionary",
+]
+
+
+def _parse_only_scripts(only_scripts):
```
samuel-duhaime, line 70: C'est bon, mais peux-tu mettre un peu de commentaire pour expliquer ce que ça fait. Surtout qu'il n'y aucun test.

packages/evolution-generator/src/scripts/generate_survey.py
samuel-duhaime: Maybe there, we could show an example of a command with this script.

samuel-duhaime (changes requested):
Je crois qu'il serait important de mettre à jour le README de evolution-generator pour montrer brièvement un exemple de cette commande.

samuel-duhaime (approved):
C'est bon!

PR #1637, opened 2026-06-16
tripDiary: add helpers to get a segment's previous/next location

tahini:
fixes #1603

Add the `fieldsWithGeojson` property to the
`SegmentSectionConfiguration` object. It contains an array that describes the various additional fields of segments that contain a value that maps to a geolocalized point. The array is sorted such that the first elements are earlier in the trip (like entry stations), while the last ones are the furthest (like exit stations). It supports fields that can be geolocation (like map widgets), or a value that maps to a feature from a collection (like select a stop/station widget).

Add helpers in the segments section to get the previous/next locations from a segment and segment's own origin/destination. The previous/next locations look at a segment's preceding/following segments and sees if any has a field set that maps to a geolocalized point. It falls back to the trip's origin/destination. The segment's own origin/destination looks at the segment itself and sees if any of the geolocalized field is set and returns this value.


packages/evolution-common/src/services/questionnaire/sections/segments/helpers.ts
```diff
@@ -203,3 +205,297 @@ export const getFilteredModesPre = (availableModes: Mode[]): ModePre[] => {
...
+ */
+const getTripDestination = ({
+    trip,
+    journey
+}: {
+    journey: Journey;
+    trip: Trip;
+}): GeoJSON.Feature<GeoJSON.Point> | null => {
+    const visitedPlaces = odHelpers.getVisitedPlaces({ journey });
+    const origin = odHelpers.getDestination({ trip, visitedPlaces });
```
kaligrafy: destination? Renommer la variable

packages/evolution-common/src/services/questionnaire/types/SectionConfig.ts
```diff
@@ -517,6 +517,27 @@ export type SegmentSectionConfiguration = {
...
+     *   the available choices for this field.
+     *
+     * The array should be ordered such that fields appearing first are closer to the
+     * segment origin, and fields appearing last are closer to the destination. This
+     * ordering informs the helpers when determining previous/next known locations
+     * during segment entry. For example, if there are multiple segments, the
+     * origin/destination of each segment will use the location data specified by
+     * these fields based on their position in the journey.
+     */
+    fieldsWithGeojson?: ({ fieldName: string } & (
```
kaligrafy: renommer fieldsWithGeojsonPoint

kaligrafy (changes requested):

kaligrafy (approved):

PR #1671, opened 2026-06-23
Provide a default Google map id

greenscientist:
The Map ID are not secret information, just a style to use for the map. Instead of asking every users to create a new one, with potential errors, we provide a default one here. It was created on the chairemobilite google project

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Google Maps configuration reliability by ensuring a valid map ID is always available, even when environment variables are not set.


tahini (approved):
Ok, si ça n'a pas d'impact sur nous, allons-y.

kaligrafy (approved):

PR #1681, opened 2026-06-25
Fix joinWith for widgets inside a group (#1621)

kaligrafy:
## Summary

Fixes #1621: `joinWith` had no effect for widgets inside a group.

- **Cause**: `SingleWidget` always read the *next* widget status from the top-level `interview.widgets`, but grouped widgets store their status under `groups.<shortname>.<id>`. The `isVisible` check therefore failed and the widgets were never joined.
- **Fix**: pass the widget status path down (`widgets` outside a group, `groups.<shortname>.<id>` inside) and read the next widget status from that same path.
- **CSS**: the dashed vertical separator is drawn as a `repeat-y` background on the label column. The join spacing previously lived on the box outer padding, leaving a 1rem gap in the dotted line between joined widgets. The spacing now lives inside the label/input columns so the separator stays continuous.

## Test plan

- [x] New parametric-style regression test in `Widget.test.tsx` that sets the next widget status only in the group path; it fails on `main` and passes with the fix.
- [x] `yarn jest Widget.test.tsx` — 19/19 pass.
- [x] Manual check in `demo_survey` (household members: Sexe joined with Occupation principale), separator continuous.


tahini (approved):
ça a l'air bon, je ne comprends rien du css, mais j'imagine que tu as testé et que c'est beau?

PR #1690, opened 2026-06-26
audits: add HH_F_AtLeastOneTransitSegmentInHousehold check

kaligrafy:
## Summary
- Port `H_F_atLeastOneTransitSegmentInHousehold` from od_mtl_2023 as `HH_F_AtLeastOneTransitSegmentInHousehold` (version 1)
- Info-level flag when any household trip has a transit segment (`trip.hasTransit()`)
- EN/FR locale strings and parametric unit tests (transit vs non-transit, empty chains)

## Test plan
- [x] `yarn jest packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts`
- [x] `yarn jest packages/evolution-backend/src/services/audits/__tests__/AuditLocales.test.ts`

## Summary by CodeRabbit

* **New Features**
  * Added a new household audit that flags when at least one trip includes a transit segment.
  * Added English and French message text for the new audit.
* **Tests**
  * Added coverage for transit vs. non-transit scenarios, including cases with missing or empty intermediate household data.

### DIFF
--- a/packages/evolution-backend/src/services/audits/auditChecks/checks/HouseholdAuditChecks.ts
+++ b/packages/evolution-backend/src/services/audits/auditChecks/checks/HouseholdAuditChecks.ts
@@ -1,5 +1,5 @@
 /*
- * Copyright 2025, Polytechnique Montreal and contributors
+ * Copyright Polytechnique Montreal and contributors
  *
  * This file is licensed under the MIT License.
  * License text available at https://opensource.org/licenses/MIT
@@ -263,6 +263,33 @@ export const householdAuditChecks: { [errorCode: string]: HouseholdAuditCheckFun
                 version: 1,
                 level: 'warning',
                 message: 'Car number per potential driving license holder > 3',
+                ignore: false
+            };
+        }
+
+        return undefined; // No audit needed
+    },
+
+    /**
+     * Info flag when the household has at least one transit segment in any trip.
+     * @param context - HouseholdAuditCheckContext
+     * @returns AuditForObject
+     */
+    HH_F_AtLeastOneTransitSegmentInHousehold: (context: HouseholdAuditCheckContext): AuditForObject | undefined => {
+        const { household } = context;
+
+        const hasTransitSegment = (household.members ?? []).some((person) =>
+            (person.journeys ?? []).some((journey) => (journey.trips ?? []).some((trip) => trip.hasTransit()))
+        );
+
+        if (hasTransitSegment) {
+            return {
+                objectType: 'household',
+                objectUuid: household.uuid!,
+                errorCode: 'HH_F_AtLeastOneTransitSegmentInHousehold',
+                version: 1,
+                level: 'info',
+                message: 'At least one transit trip in household',
                 ignore: false
             };
         }
--- a/packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
+++ b/packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
@@ -1,0 +1,89 @@
+/*
+ * Copyright Polytechnique Montreal and contributors
+ *
+ * This file is licensed under the MIT License.
+ * License text available at https://opensource.org/licenses/MIT
+ */
+
+import { v4 as uuidV4 } from 'uuid';
+import { householdAuditChecks } from '../../HouseholdAuditChecks';
+import { createContextWithHouseholdAndHome } from './testHelper';
+import { Person } from 'evolution-common/lib/services/baseObjects/Person';
+import { Journey } from 'evolution-common/lib/services/baseObjects/Journey';
+import { Trip } from 'evolution-common/lib/services/baseObjects/Trip';
+import { Segment } from 'evolution-common/lib/services/baseObjects/Segment';
+import { Mode } from 'evolution-common/lib/services/baseObjects/attributeTypes/SegmentAttributes';
+import { SurveyObjectsRegistry } from 'evolution-common/lib/services/baseObjects/SurveyObjectsRegistry';
+
+describe('HH_F_AtLeastOneTransitSegmentInHousehold audit check', () => {
+    const validHouseholdUuid = uuidV4();
+    const validHomeUuid = uuidV4();
+    const surveyObjectsRegistry = new SurveyObjectsRegistry();
+
+    const expectedInfoAudit = {
+        objectType: 'household',
+        objectUuid: validHouseholdUuid,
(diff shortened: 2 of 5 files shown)


packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
```diff
@@ -0,0 +1,142 @@
...
+import { SurveyObjectsRegistry } from 'evolution-common/lib/services/baseObjects/SurveyObjectsRegistry';
+
+type ArrayState = 'undefined' | 'empty' | 'present';
+
+describe('HH_F_AtLeastOneTransitSegmentInHousehold audit check', () => {
+    const validHouseholdUuid = uuidV4();
+    const validHomeUuid = uuidV4();
+    const surveyObjectsRegistry = new SurveyObjectsRegistry();
+
+    const resolveArray = <T>(state: ArrayState, items: T[]): T[] | undefined => {
```
tahini: Cette fonction mérite une description: `resolveArray`?

kaligrafy: ajouté

packages/evolution-backend/src/services/audits/auditChecks/checks/HouseholdAuditChecks.ts
```diff
@@ -221,6 +221,33 @@ export const householdAuditChecks: { [errorCode: string]: HouseholdAuditCheckFun
...
+        const { household } = context;
+
+        const hasTransitSegment = (household.members ?? []).some((person) =>
+            (person.journeys ?? []).some((journey) => (journey.trips ?? []).some((trip) => trip.hasTransit()))
+        );
+
+        if (hasTransitSegment) {
+            return {
+                objectType: 'household',
+                objectUuid: household._uuid!,
```
greenscientist: Just wondering, why is the uuid marked as "private" with the underscore and we use it here?

greenscientist (changes requested):
Does the tests code need to be that complexe?  Seems like to went really out of our way to make a single parametric test. 

I just have the feel that this is a prompt gone wild and would have been better with more explicit test case. The resolveArray thing is weird. 

So if you really want to keep it that way, I would need a lenghty justification in the commit message that feels like it's written by you and not by Claude. 

To give you an hint, your test seem tu always have just one element that is "empty"/undefined, so instead of having something to build a complex chain, it feels like you can have fully define object where you can delete just one element for each test.

packages/evolution-backend/src/services/audits/auditChecks/checks/HouseholdAuditChecks.ts
kaligrafy: Good catch — fixed, now uses the public `.uuid` getter (defined on `Uuidable`) instead of reaching into `_uuid` directly. That's the correct public API; the underscore is just the backing field convention used throughout the base object classes.

greenscientist (changes requested):
You need to squash your commit

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
```diff
@@ -0,0 +1,89 @@
...
+        const result = runCheck([makePersonWithSegmentMode('transitBus')]);
+        expect(result).toMatchObject(expectedInfoAudit);
+    });
+
+    it('does not flag a household with only non-transit segments', () => {
+        const result = runCheck([makePersonWithSegmentMode('walk'), makePersonWithSegmentMode('carDriver')]);
+        expect(result).toBeUndefined();
+    });
+
+    it.each([
```
greenscientist, line 72: It's a bit weird to have a small parametrized test for only 2 simple element. I think it would be clearer at that point to keep them as separate.

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        const journey = new Journey({ _uuid: uuidV4() }, surveyObjectsRegistry);
+        journey.trips = [trip];
+
+        const person = new Person({ _uuid: uuidV4() }, surveyObjectsRegistry);
+        person.journeys = [journey];
+        return person;
+    };
+
+    const runCheck = (members: Person[] | undefined) => {
```
greenscientist: This need a description. 

Also you should review if you really want a function for a essentialy a one liner (Could be fine, but you might want a more descriptive name than "runCheck"

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/testHelper.ts
```diff
@@ -16,6 +16,9 @@ import { createMockHome } from '../home/testHelper';
 export const createMockHousehold = (overrides: Partial<Household> = {}, validUuid = uuidV4()) => {
     return {
         _uuid: validUuid,
+        get uuid() {
```
greenscientist, line 19: Is get uuid part of the Household interface ?? You check that it match. Or add a comment to explain that it's an extra field.

greenscientist (approved):
I'm accepting, with reserve. I still have a few comment, it still feel like the AI is trying too much to look clever.

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/testHelper.ts
kaligrafy: it is part of the survey objects interfaces

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
kaligrafy: Ça ça vient de @tahini ;-P

packages/evolution-backend/src/services/audits/auditChecks/checks/__tests__/household/HH_F_AtLeastOneTransitSegmentInHousehold.test.ts
kaligrafy: renommé runAtLeastOneTransitSegmentCheck pour plus de précision

PR #1691, opened 2026-06-26
Update changelog and comments now that the Google Map ID is optional

greenscientist:
In commit a21a298 we added a default Map ID, so we don't have to add a new one in each setup. This update the changelog and the .env.example to reflect the fact that the environmment variable is now optional

## Summary by CodeRabbit

* **Documentation**
  * Updated the changelog entry to clarify that an optional Google Map ID environment setting can override the default map style, without the previous deployment/rendering caveats.
  * Revised the sample environment file to make the Google Map ID override commented out by default, with updated guidance for when and how to use it.


tahini (approved):
C'est bon, mais il y a des conflits et erreurs de grammaire à corriger

kaligrafy (approved):

tahini (approved):
C,est bon, les changemetns n'ont pas besoin d'être breaking, c'est à la bonne place

PR #1742, opened 2026-07-22
build(deps): update Transition and align dependency versions

kaligrafy:
## Summary
- Bump Transition submodule to `18506578`, picking up Dependabot security updates merged on Transition main (body-parser, engine.io, immutable, fast-uri, js-yaml, etc. via chaire-lib).
- Bump Evolution-specific packages: `dompurify` ^3.4.12, `morgan` ^1.11.0, `multer` ^2.2.0.
- Regenerate `yarn.lock` and run `yarn-deduplicate` to consolidate duplicate lockfile entries.

## Test plan
- [x] `yarn compile`
- [x] Manual smoke test: admin UI (`yarn start:admin`)
- [x] Manual smoke test: participant questionnaire

## Summary by CodeRabbit

* **Chores**
  * Updated backend and frontend dependencies to newer versions, including security-related sanitization, HTTP request logging, and file upload libraries.
  * Refreshed the bundled transition component to a more recent revision to keep the packaged experience current.


greenscientist (changes requested):
You should squash the commit, in both case you say you ran yarn deduplicate, I don't see the point of the second commit.

greenscientist (approved):

PR #1745, opened 2026-07-23
refactor(person-factory): align Person.create call with other factories

kaligrafy:
## Summary

PersonFactory was using an old create pattern, we need to update it to be consistent with other base objects factories.

- Omit nested `journeys` before calling `Person.create`, since journeys are built separately by `JourneyFactory`
- Cast person attributes to `ExtendedPersonAttributes`, matching the pattern used in `JourneyFactory` and `TripFactory`
- Add a unit test verifying `journeys` are not passed to `Person.create`

## Test plan
- [x] `yarn test --testPathPatterns='PersonFactory.test.ts'` passes (7 tests)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved survey object creation by populating each person’s journeys after successful creation.
  * Ensured work and school locations are configured after journey population.
  * Prevented follow-up setup when person creation fails, reducing incomplete survey data.

* **Tests**
  * Expanded coverage for person creation, journey population, and work/school location setup.
  * Updated household creation tests to verify the revised processing flow.

### DIFF
--- a/packages/evolution-backend/src/services/surveyObjects/PersonFactory.ts
+++ b/packages/evolution-backend/src/services/surveyObjects/PersonFactory.ts
@@ -5,20 +5,27 @@
  * License text available at https://opensource.org/licenses/MIT
  */
 
+import _omit from 'lodash/omit';
+
 import { SurveyObjectsWithErrors } from 'evolution-common/lib/services/baseObjects/types';
 import { CorrectedResponse } from 'evolution-common/lib/services/questionnaire/types';
 import { Person } from 'evolution-common/lib/services/baseObjects/Person';
 import { isOk } from 'evolution-common/lib/types/Result.type';
 import { Household } from 'evolution-common/lib/services/baseObjects/Household';
+import { Home } from 'evolution-common/lib/services/baseObjects/Home';
+import { Optional } from 'evolution-common/lib/types/Optional.type';
 import projectConfig from '../../config/projectConfig';
 import { SurveyObjectsRegistry } from 'evolution-common/lib/services/baseObjects/SurveyObjectsRegistry';
 import { compareSequenceThenUuid } from 'evolution-common/lib/services/baseObjects/sequenceUtils';
+import { populateJourneysForPerson } from './JourneyFactory';
+import { ExtendedPersonAttributes } from 'evolution-common/lib/services/baseObjects/Person';
 
 /**
  * Generate persons
  * Populate members for a household from the household's persons attributes
  * @param {SurveyObjectsWithErrors} surveyObjectsWithErrors - Container for created objects with errors
  * @param {Household} household - The household to add the members to
+ * @param {Home} home - The home object for geography assignment, needed by nested journeys
  * @param {CorrectedResponse} correctedResponse - corrected response
  * @param {SurveyObjectsRegistry} surveyObjectsRegistry - SurveyObjectsRegistry
  * @returns {Promise<void>}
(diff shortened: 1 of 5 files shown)


greenscientist (changes requested):
Please explain why you need to omit nested journeys. (At least in the commit message, but it probably need a comment in the code also.

packages/evolution-backend/src/services/surveyObjects/PersonFactory.ts
```diff
@@ -59,7 +61,10 @@ export async function populatePersonsForHousehold(
             ? projectConfig.surveyObjectParsers.person(originalCorrectedPersonAttributes, correctedResponse)
             : originalCorrectedPersonAttributes;
 
-        const personResult = Person.create(personAttributes as { [key: string]: unknown }, surveyObjectsRegistry);
+        const personResult = Person.create(
```
tahini, line 68: Les autres classes qui utilisent ce pattern avec le `_omit` ajoutent un appel du genre `populateVisitedPlacesForJourney` dans le bloc de `isOk`, il ne faudrait pas faire ça ici aussi, en ajoutant le `populateJourneysForPerson`?

tahini (changes requested):

packages/evolution-backend/src/services/surveyObjects/__tests__/SurveyObjectsFactory.test.ts
```diff
@@ -114,25 +110,16 @@ describe('SurveyObjectsFactory', () => {
             expect(result.home).toBe(mockHome);
             expect(result.household).toBe(mockHousehold);
 
-            // Verify person and journey factories were called
+            // Verify person factory was called (journeys/setupWorkAndSchoolPlaces are now
+            // triggered from within populatePersonsForHousehold, covered by PersonFactory.test.ts)
```
tahini: On n'a pas besoin de savoir ça, "are now triggered", now par rapport à quoi? Ceci est un commentaire justifiant le changement (genre dans le commit message), ça ne va pas dans le code

packages/evolution-backend/src/services/surveyObjects/PersonFactory.ts
```diff
@@ -59,13 +66,34 @@ export async function populatePersonsForHousehold(
             ? projectConfig.surveyObjectParsers.person(originalCorrectedPersonAttributes, correctedResponse)
             : originalCorrectedPersonAttributes;
 
-        const personResult = Person.create(personAttributes as { [key: string]: unknown }, surveyObjectsRegistry);
+        // journeys is omitted here for the same reason visitedPlaces/trips are omitted before
+        // Journey.create in JourneyFactory: it is a nested child collection, not a Person
+        // attribute. Once the person itself is successfully created, populateJourneysForPerson
+        // builds each journey individually so that per-journey errors are tracked separately
+        // instead of failing the whole Person.create call.
```
tahini: Svp, revoir les commentaires écrits par Claude, souvent, ils ne sont pas pertinents dans le code et ne font qu'expliquer pouquoi le changement a été fait. Il faut lire ce commentaire du point du vue de la personne qui va lire le code et vouloir le modifier dans 6 mois. Ici, simplement `// Omit journeys as they will be populated separately in the next step` est suffisant, pas besoin de comparer avec le visitedPlaces/trips car ça veut dire que le commentaire ici dépend de l'implémentation là-bas, il ne faut pas.

tahini (approved):

greenscientist (approved):
I guess I'll accept. 
I feel that the omitting should be handled by the create function directly,  maybe we some flags passed to create to tell it to omit certain part. 
Or maybe we should pass new journeys directly to  that create function. 
Just some thoughts for your considerations.

PR #1791, opened 2026-08-06
fix(generator): parse InputRange numeric cells in questionnaire dicti…

kaligrafy:
…onary

Handle string and float min/max values from Excel and skip incomplete InputRange rows without raising when minValue is empty.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of questionnaire ranges with numeric values, numeric text, and empty cells.
  * Preserved missing minimum values instead of incorrectly assigning defaults.
  * Rejected boolean values where integer ranges are expected.

* **Tests**
  * Expanded coverage for numeric parsing, incomplete ranges, mixed input types, and generated range labels.

### DIFF
--- a/packages/evolution-generator/src/scripts/generate_questionnaire_dictionary.py
+++ b/packages/evolution-generator/src/scripts/generate_questionnaire_dictionary.py
@@ -10,6 +10,19 @@ from typing import Literal
 from helpers.generator_helpers import get_data_from_excel, clean_text
 
 
+def parse_int_cell_value(value):
+    """Parse an Excel cell value as an integer (handles numeric strings)."""
+    if value is None or value == "":
+        return None
+    if isinstance(value, bool):
+        raise TypeError(f"Expected a numeric cell value, got boolean: {value}")
+    if isinstance(value, int):
+        return value
+    if isinstance(value, float):
+        return int(value)
+    return int(float(str(value).strip()))
+
+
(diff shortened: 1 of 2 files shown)


packages/evolution-generator/src/scripts/generate_questionnaire_dictionary.py
```diff
@@ -311,9 +324,12 @@ def process_range(ranges_rows, ranges_headers, language):
     ranges_map = {}
     for row in ranges_rows[1:]:
         input_range_name = row[input_range_name_index].value
-        # Ensure min_value is not negative
-        min_value = max(0, row[min_value_index].value)
-        max_value = row[max_value_index].value
+        parsed_min_value = parse_int_cell_value(row[min_value_index].value)
+        # Ensure min_value is not negative, but keep None for incomplete rows
+        min_value = (
```
tahini, line 329: It is not the dictionary's job to do this. We have to make sure that whatever widget data is generated is inline with the dictionary. If we really want the min to be non-negative (is it the case? is it documented?), then we would need a helper function that makes sure of that and is called both by the dictionary and widget generation. In the present case, we create a possible discrepancies between both. Granted, it was already the case and you are fixing a bug, but then, the bug would also occur in the widget, no? Or if it,s already fixed in the widget, use the same function here.

tahini (changes requested):

PR #1820, opened 2026-08-14
fix(validations): show required error first for blank household asset…

kaligrafy:
… counts

Put the _isBlank check before the numeric format rule so a missing count fails as required, as requested on #1814.

### DIFF
--- a/packages/evolution-common/src/services/widgets/validations/householdAssetCountValidation.ts
+++ b/packages/evolution-common/src/services/widgets/validations/householdAssetCountValidation.ts
@@ -82,19 +82,19 @@ const createHouseholdAssetCountValidation = (
 
         return [
             {
-                validation: !_isBlank(value) && (isNaN(Number(value)) || !Number.isInteger(Number(value))),
-                errorMessage: (t) => t(messageKeys.invalid)
-            },
-            {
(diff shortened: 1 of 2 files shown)


packages/evolution-common/src/services/widgets/validations/householdAssetCountValidation.ts
```diff
@@ -81,20 +81,20 @@ const createHouseholdAssetCountValidation = (
...
                 validation: _isBlank(value),
                 errorMessage: (t) => t(messageKeys.required)
             },
             {
-                validation: Number(value) < 0,
+                validation: !_isBlank(value) && (isNaN(Number(value)) || !Number.isInteger(Number(value))),
+                errorMessage: (t) => t(messageKeys.invalid)
+            },
+            {
+                validation: !_isBlank(value) && (isNaN(Number(value)) || Number(value) < 0),
```
tahini: Il ne faudrait pas plutôt avoir !isNaN && < 0? car NaN va retourner les 3 derniers, non?

kaligrafy: corrigé

tahini (approved):

Issue #1848, opened 2026-08-20
section templates: the activity name is not escaped or sanitized in section templates


tahini:
This could be a place for xss injection if the participant enter faulty javascript as place name

tahini:
ftr, there is no injection possible, it just doesn't look good without prior sanitation of characters.

PR #1849, opened 2026-08-20
labels: sanitize strings at render instead of escaping

### LINKED ISSUE #1838: labels: escaped caracters sometimes appear in labels
See the following screenshot, also happens in other fields

Image

tahini:
fixes #1838

The OD helper functions do not escape the participant data anymore, as
this causes places showing the data as plain text to display escaped
HTML characters.

It is the role of the renderers to correctly sanitize strings coming
from the participant, or the app in general. Where plain text is used,
escaping is handled naturally by `React`. It is where the
`dangerouslySetInnerHtml` is used to display strings from the user that
there may be problems. So in those calls, we make sure the call the
`stripUnsafeHtml` function to remove any undesired HTML markup that may
have been entered by the participants.

## Summary by CodeRabbit

* **Security**
  * Improved HTML sanitization across surveys, forms, maps, modals, labels, and error messages.
  * Unsafe markup and links are removed while approved formatting remains available where appropriate.
* **Bug Fixes**
  * Survey text, participant details, addresses, and visited-place names now display consistently without unwanted escaping.
  * Plain-text contexts no longer show HTML markup.
* **Documentation**
  * Updated the unreleased changelog with the new rendering and sanitization behavior.


tahini:
This PR is based on #1847 and will be rebased once this one is merged.

I have another PR in the queue for the VisitedPlaces and TripsAndSegments template, to fix the escaping there too as part of #1848

kaligrafy (approved):

PR #1854, opened 2026-08-20
Release/0.6.0

tahini:
Tag this commit to 0.6.0 version


tahini:
This is the version that will be used for the start of the next survey cycles. We will do frequent intermediate releases as development continues and bugs are fixed, no need to branch out yet, but there should be no breaking changes from now on in the 0.6.0 release cycle.

This approach will allow to better track versions and make updates easier on surveys, with fixed version instead of arbitrary state of the `main` branch in supported surveys.

kaligrafy (approved):

